PR #17, opened 2019-03-05
Handle leading zeros in (r|s)-parameters of ECDSA signature

xbill-dns:
As per https://github.com/ibauersachs/dnssecjava/issues/14

### DIFF
--- a/org/xbill/DNS/DNSSEC.java
+++ b/org/xbill/DNS/DNSSEC.java
@@ -775,42 +775,86 @@ ECGOSTSignaturefromDNS(byte [] signature, ECKeyInfo keyinfo)
 	return signature;
 }
 
-private static byte []
-ECDSASignaturefromDNS(byte [] signature, ECKeyInfo keyinfo)
-	throws DNSSECException, IOException
-{
-	if (signature.length != keyinfo.length * 2)
-		throw new SignatureVerificationException();
+/**
+ * 
+ * Convert a DNS standard ECDSA signature (defined in RFC 6605) into a JCE
+ * standard ECDSA signature, which is encoded in ASN.1.
+ * 
+ * The format of the ASN.1 signature is
+ * 
+ * ASN1_SEQ . seq_length . ASN1_INT . r_length . R . ANS1_INT . s_length . S
+ * 
+ * where R and S may have a leading zero byte if without it the values would be
+ * negative.
+ *
+ * The format of the DNSSEC signature is just R . S where R and S are both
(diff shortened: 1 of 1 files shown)


org/xbill/DNS/DNSSEC.java
```diff
@@ -869,12 +913,10 @@ private Algorithm() {}
...
 				break;
 			case Algorithm.ECDSAP256SHA256:
-				signature = ECDSASignaturefromDNS(signature,
-								  ECDSA_P256);
+				signature = convertECDSASignature(signature);
 				break;
 			case Algorithm.ECDSAP384SHA384:
-				signature = ECDSASignaturefromDNS(signature,
-								  ECDSA_P384);
+				signature = convertECDSASignature(signature);
```
kingle: Duplicate branches in switch.  If both this case and the one above it are to call the same exact method, would it would be cleaner to:
```
  case Algorithm.ECDSAP256SHA256:
  case Algorithm.ECDSAP384SHA384:
    signature = convertECDSASignature(signature);
    break;
```

org/xbill/DNS/DNSSEC.java
```diff
@@ -775,42 +775,86 @@ private Algorithm() {}
...
+ * The format of the DNSSEC signature is just R . S where R and S are both
+ * exactly "length" bytes.
+ * 
+ * @param signature
+ *            The binary signature data from an RRSIG record.
+ * @return signature data that may be used in a JCE Signature object for
+ *         verification purposes.
+ */
+public static byte[] convertECDSASignature(byte[] signature) {
+	/**
```
kingle: Dangling javadoc.  Perhaps just use the regular block comment?

org/xbill/DNS/DNSSEC.java
```diff
@@ -775,42 +775,86 @@ private Algorithm() {}
...
+	 * This method is a patch for the problem of leading zeros that keep the the
+	 * public key from being verified even if it is valid
+	 * See: https://github.com/ibauersachs/dnssecjava/issues/14
+	 * 
+	 * This patch was obtained from:
+	 * https://github.com/dblacka/jdnssec-tools/pull/4/commits/ca2a9324858b6b5a87fde8ce8a5fb2365c85db35
+	 * and
+	 * https://github.com/gryphius/dnsjava/commit/b4216b35a8386cf47e2c59a875c655d584624d1d
+	 */
+	{
```
kingle: Unneeded extra braces {}  ?

PR #23, opened 2019-05-18
CDS/CDNSKEY Record Types

SoylentBob:


ibauersachs:
Thanks!

I'll need some time to do a proper review. In the meantime, could you please clean this up so that it only contains the actual code changes and no formatting changes? I know that dnsjava has a rather atypical style, but to preserve a clean history I'd like to keep it.

SoylentBob:
I think I got rid of the biggest formatting changes now. Let me know if there is any more stuff that you like to be changed.

ibauersachs:
Thanks! But actually, yes, there are some more things:
- Please use regular Javadoc headers (i.e. just two asterisks)
- Braces on methods go on the same line as the declaration, except if the declaration itself is multi-line
- How about `DNSKEYRecordBase` and `DSRecordBase`? This would be more consistent with `KEYBase`
- There are still changes in DNSKEYRecord, DSRecord, Type that are not necessary

ibauersachs:
Ahm, I've just scrolled over RFC 7344 and saw that the CDS / CDNSKEY records are actually identical to the regular DNSSEC records. Instead of extracting all functionality to a base class and break existing APIs, I think it would make more sense - and give much less changes - to:
- create two new classes for the new records
- derive from the DNSSEC class respectively
- Add a protected constructor to the DNSSEC classes allowing to specify the `Type`

PR #25, opened 2019-05-18
Ephemeral port

ibauersachs:


org/xbill/DNS/UDPClient.java
```diff
@@ -83,13 +83,17 @@ public void run() {
...
 void
 bind(SocketAddress addr) throws IOException {
-	if (addr == null ||
-	    (addr instanceof InetSocketAddress &&
-	     ((InetSocketAddress)addr).getPort() == 0))
-	{
-		bind_random((InetSocketAddress) addr);
-		if (bound)
-			return;
+	boolean useEphemeralPort = Boolean.parseBoolean(System.getProperty(UDPCLIENT_DEFAULT_TO_EPHEMERAL_PORT, "false"));
```
kentros: nit: `parseBoolean` was introduced in Java 1.5.  Is there a plan to go forward with bumping the minimum target?

ibauersachs: This is just a backup of a patch from Sourceforge. But yes, I'll update the target to 1.5 as the code already uses some 1.5 stuff.

PR #28, opened 2019-05-18
Target Java 1.5. fix #1

### LINKED ISSUE #1: Bump up default target.jdk
Any chance of bumping the default target.jdk to something like 1.6 (or even 1.7) instead of 1.4?  I ask because it causes some minor issues when using within Android if the class files are an old binary level.

kentros:
Given we're already using code that requires Java 1.5 and it makes sense to target it now.  Bumped major version because [semver](https://semver.org/).

Also including [animal-sniffer-plugin](https://www.mojohaus.org/animal-sniffer/animal-sniffer-maven-plugin/) to fail builds if they include code outside the target.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,12 +43,12 @@ still exists.
 
 ## Getting started
 
-Run `mvn package` from the toplevel directory to build dnsjava. JDK 1.4
+Run `mvn package` from the toplevel directory to build dnsjava. JDK 1.5
 or higher is required.
 
 ### Replacing the standard Java DNS functionality:
 
-Java versions from 1.4 to 1.8 can load DNS service providers at runtime. The
+Java versions from 1.5 to 1.8 can load DNS service providers at runtime. The
 functionality was [removed in JDK 9](https://bugs.openjdk.java.net/browse/JDK-8134577),
 a replacement is [requested](https://bugs.openjdk.java.net/browse/JDK-8192780),
 but so far has not been implemented.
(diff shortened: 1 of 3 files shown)


ibauersachs:
I don't want to change to v3 just yet. What are your thoughts to my comment in #1?

README.md
```diff
@@ -43,12 +43,12 @@ still exists.
...
 ## Getting started
 
-Run `mvn package` from the toplevel directory to build dnsjava. JDK 1.4
+Run `mvn package` from the toplevel directory to build dnsjava. JDK 1.5
 or higher is required.
 
 ### Replacing the standard Java DNS functionality:
 
-Java versions from 1.4 to 1.8 can load DNS service providers at runtime. The
+Java versions from 1.5 to 1.8 can load DNS service providers at runtime. The
```
ibauersachs: This one here is actually correct as it is a historical reference to 1.4.

build.xml
```diff
@@ -5,26 +5,26 @@
     <property name="dist_dir" value="${basedir}"/>
     <property name="tests_dir" value="${basedir}/tests"/>
     <property name="reports_dir" value="${basedir}/reports"/>
-    <property name="version" value="2.1.8"/>
+    <property name="version" value="3.0.0"/>
```
kentros: Do we still want to keep this build.xml around?  Can it be removed now that we're using maven?

ibauersachs: Yes for 2.1.9, no for anything that comes afterwards.

ibauersachs: And speaking of it, after 2.1.9, I'll clean the file/directory layout to standard `src/{main,test}/java`.

kentros:
Closing this out until the new layout is in place

PR #36, opened 2019-05-21
Adds japicmp plugin to ensure compatibility

kingle:
Per comment in https://github.com/dnsjava/dnsjava/issues/1#issuecomment-493774420
> when I see how much impact the other API changes will have. E.g. adding generics will not necessarily break APIs.

+1 on ensuring APIs don't break.  With this plugin, the build will fail when it detects incompatible API changes, compared to the previous version.  Also gives some insight/sanity checks on non-breaking changes (adding to API, removing method identical to superclass' implementation, etc.).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -149,6 +149,25 @@
                 <artifactId>coveralls-maven-plugin</artifactId>
                 <version>4.3.0</version>
             </plugin>
+
+            <plugin>
+                <groupId>org.codehaus.mojo</groupId>
+                <artifactId>clirr-maven-plugin</artifactId>
+                <version>2.8</version>
+                <configuration>
+                    <!-- JDK9 removed the Nameservice SPI as per bug 8134577. -->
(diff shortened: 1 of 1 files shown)


ibauersachs:
I've previously used japicmp-maven-plugin in jitsi/ice4j and rejected clirr-maven-plugin, but can't remember why atm. But it seems weird that you don't need to specify a baseline version. What does clirr use?

kingle:
japicmp-maven-plugin looks better.  I'll switch to use it -- according to their site, clirr is basically dead: https://siom79.github.io/japicmp/

kingle:
Switched it with japicmp -- let me know if you prefer other parameter config: https://siom79.github.io/japicmp/MavenPlugin.html

kingle:
baseline (old version) is latest released version if not specified.

ibauersachs:
Let's see how this works out.

PR #37, opened 2019-05-22
Miscount in Message.sectionToWire

demonti:
This is the patch for the fix of the issue #60 "Miscount in Message.sectionToWire" in the old SourceForge issue system I opened back then.


ibauersachs:
https://sourceforge.net/p/dnsjava/bugs/60/

ibauersachs:
Thanks! Would you mind adding a unit test that checks the particular case with the OPTRecord?

demonti:
I have added a respective test case. It fails on the current code and succeeds on the fix.

PR #40, opened 2019-05-24
added fix and test case for the SourceForge issue #59

demonti:
issue #59 "Record.getRRsetType() causes ClassCastException"


demonti:
Hmm, I do not understand the barking of coveralls…

ibauersachs:
> Hmm, I do not understand the barking of coveralls…

Me neither, just ignore it. I think some of the tests still run queries against external servers, this might cause different code paths to be taken. It's more meant as a general indication.

https://sourceforge.net/p/dnsjava/bugs/59/

PR #42, opened 2019-05-27
Introduce generics

kingle:
First pass to add generics.  Most of the work was done automatically from the [IntelliJ Generify tool](https://www.jetbrains.com/help/idea/generify-refactoring.html)

Some classes may need some more TLC if we're to generify them, but most are internal types.  The following classes have some parts that were intentionally not part of this PR because the changes may be more invasive:

- Cache and Zone both have an internal Map `data` with dynamic class values (sometimes lists, sometimes not)
- Message has an internal array `sections` with dynamic List classes

### DIFF
--- a/src/main/java/org/xbill/DNS/RRset.java
+++ b/src/main/java/org/xbill/DNS/RRset.java
@@ -22,14 +22,14 @@ private static final long serialVersionUID = -3270249290171239695L;
  * rrs contains both normal and RRSIG records, with the RRSIG records
  * at the end.
  */
-private List rrs;
+private ArrayList<Record> rrs;
 private short nsigs;
 private short position;
 
 /** Creates an empty RRset */
 public
 RRset() {
-	rrs = new ArrayList(1);
+	rrs = new ArrayList<>(1);
 	nsigs = 0;
 	position = 0;
 }
@@ -45,7 +45,7 @@ RRset(Record record) {
 public
 RRset(RRset rrset) {
 	synchronized (rrset) {
-		rrs = (List) ((ArrayList)rrset.rrs).clone();
+		rrs = new ArrayList<>(rrset.rrs);
 		nsigs = rrset.nsigs;
 		position = rrset.position;
 	}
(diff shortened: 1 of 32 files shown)


ibauersachs:
Uhm, I basically did more or less the same yesterday :-/
Please let me know beforehand if you're working on something so we don't do it twice again.

src/main/java/org/xbill/DNS/RRset.java
```diff
@@ -160,13 +160,13 @@
...
  */
-public synchronized Iterator
+public synchronized Iterator<Record>
 rrs() {
 	return iterator(true, true);
 }
 
 /** Returns an Iterator listing all signature records */
-public synchronized Iterator
+public synchronized Iterator<Record>
```
ibauersachs: These are all RRSIGRecords and can thus be returned like that. The casts are not source-compatible anyway if used like e.g. https://github.com/ibauersachs/dnssecjava/blob/master/src/main/java/org/jitsi/dnssec/validator/ValUtils.java#L404

kingle:
> Uhm, I basically did more or less the same yesterday :-/
Please let me know beforehand if you're working on something so we don't do it twice again.

Will do.  Want me to close this PR out?

src/test/java/org/xbill/DNS/SetResponseTest.java
```diff
@@ -77,21 +70,13 @@ void test_ctor_1arg()
     @Test
     void test_ctor_1arg_toosmall()
     {
-	try {
-	    new SetResponse(-1);
-	    fail("IllegalArgumentException not thrown");
-	}
-	catch(IllegalArgumentException a){}
+    	assertThrows(IllegalArgumentException.class, () -> new SetResponse(-1));
```
ibauersachs: There's a lot more to cleanup like this (unnecessary boxing, StringBuffer -> StringBuilder (already done, just not pushed yet)), but please let us do it at least _distinct_ commits.

ibauersachs:
> > Uhm, I basically did more or less the same yesterday :-/
> > Please let me know beforehand if you're working on something so we don't do it twice again.
> 
> Will do. Want me to close this PR out?

No. We'll leave it open so that if anybody else walks along sees that work is being done.

ibauersachs:
I have other changes lying around, I'll merge this one if you remove the changes the changes in SetResponseTest (they can come in another PR).

kingle:
SetResponseTest changes reverted and resolved conflicts from master

ibauersachs:
Thanks!

PR #45, opened 2019-05-30
Fix #44 Prevent SOA serial being set to zero

### LINKED ISSUE #44: Serial must not be zero
```Serial.increment()``` may return zero as seen here 
https://github.com/dnsjava/dnsjava/blob/6d50d8cc03e412642f28a0890c626f1283b8a86e/src/main/java/org/xbill/DNS/Serial.java#L57

But according to RFC1982 section 7 https://tools.ietf.org/html/rfc1982
> Caution should also be exercised before causing the serial number to
   be set to the value zero.  While this value is not in any way special
   in serial number arithmetic, or to the DNS SOA serial number, many
   DNS implementations have incorrectly treated zero as a special case,
   with special properties, and unusual behaviour may be expected if
   zero is used as a DNS SOA serial number.

and RFC2136 section 4.2 & 7.11 https://tools.ietf.org/html/rfc2136
> 4.2. If update ordering is desired, the requestor will need to know
   the value of the existing SOA RR.  Requestors who update the SOA RR
   must update the SOA SERIAL field in a positive direction (as defined
   by [RFC1982]) and also preserve the other SOA fields unless the
   requestor's explicit intent is to change them.  The SOA SERIAL field
   must never be set to zero (0).
   ...
   7.11. A zone's SOA SERIAL should never be set to zero (0) due to
   interoperability problems with some older but widely installed
   implementations of DNS.  When incrementing an SOA SERIAL, if the
   result of the increment is zero (0) (as will be true when wrapping
   around 2**32), it is necessary to increment it again or set it to one
   (1).  See [RFC1982] for more detail on this subject.

### PR DESCRIPTION
**Pull request by kingle**
RFC 2136 seems clear https://tools.ietf.org/html/rfc2136

> 7.11. A zone's SOA SERIAL should never be set to zero (0) ... When incrementing an SOA SERIAL, if the result of the increment is zero (0) (as will be true when wrapping around 2**32), it is necessary to increment it again or set it to one (1).

### DIFF
--- a/src/main/java/org/xbill/DNS/Serial.java
+++ b/src/main/java/org/xbill/DNS/Serial.java
@@ -5,11 +5,10 @@ package org.xbill.DNS;
 /**
  * Helper functions for doing serial arithmetic.  These should be used when
  * setting/checking SOA serial numbers.  SOA serial number arithmetic is
- * defined in RFC 1982.
+ * defined in RFC 1982 and also referenced in RFC 2136.
  *
  * @author Brian Wellington
  */
-
 public final class Serial {
 
 private static final long MAX32 = 0xFFFFFFFFL;
@@ -34,17 +33,14 @@ compare(long serial1, long serial2) {
 	if (serial2 < 0 || serial2 > MAX32)
 		throw new IllegalArgumentException(serial2 + " out of range");
 	long diff = serial1 - serial2;
-	if (diff >= MAX32)
-		diff -= (MAX32 + 1);
-	else if (diff < -MAX32)
-		diff += (MAX32 + 1);
 	return (int)diff;
 }
 
 /**
  * Increments a serial number.  The number is assumed to be a 32 bit unsigned
  * integer stored in a long.  This basically adds 1 and resets the value to
- * 0 if it is 2^32.
+ * 1 if the serial is 2^32.  A zero value is explicitly forbidden as per
+ * RFC 2136 section 4.2 & 7.11.
  * @param serial The serial number
  * @return The incremented serial number
  * @throws IllegalArgumentException serial is out of range
@@ -54,7 +50,7 @@ increment(long serial) {
 	if (serial < 0 || serial > MAX32)
 		throw new IllegalArgumentException(serial + " out of range");
 	if (serial == MAX32)
-		return 0;
+		return 1;
 	return serial + 1;
 }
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/main/java/org/xbill/DNS/Serial.java [hunk 1]
```diff
@@ -34,17 +33,14 @@
 	if (serial2 < 0 || serial2 > MAX32)
 		throw new IllegalArgumentException(serial2 + " out of range");
 	long diff = serial1 - serial2;
-	if (diff >= MAX32)
```
kingle: This if-elseif clause appeared to be redundant, but please sanity check me.  The long->int conversion seems to provide the exact same values as we currently check for in the SerialTest test_compare_boundary
https://github.com/dnsjava/dnsjava/blob/6d50d8cc03e412642f28a0890c626f1283b8a86e/src/test/java/org/xbill/DNS/SerialTest.java#L129

bwelling:
I don't think that this is the right fix.  The Serial class is implementing serial number arithmetic, as defined in RFC 1982.  RFC 2136 is not defining serial number arithmetic; it's defining a property of zone serial numbers.  Other uses of serial number arithmetic, if there are any, may not treat 0 as special.

If there is code implementing the server portion of RFC 2136, it should be updated to avoid setting the zone serial number to 0, presumably by either incrementing it a second time or setting it to 1.  I don't think there is any such code in dnsjava proper, though, and I'm not seeing any calls to Serial.increment outside of tests.

bwelling:
To be a bit more clear, I don't think there's anything to fix here.  The original report is describing a problem in theoretical naive code using Serial.increment() when processing dynamic updates, but that code isn't part of dnsjava.

kingle:
Disregarding RFC 2136, and sticking to RFC 1982, are we sufficiently exercising caution?

RFC1982 section 7 https://tools.ietf.org/html/rfc1982

> Caution should also be exercised before causing the serial number to
be set to the value zero. ... unusual behaviour may be expected if
zero is used as a DNS SOA serial number.

bwelling:
I'm not completely sure why I added Serial.increment (in 2003), other than the fact that it seemed potentially useful.  Perhaps it should just be removed.

src/main/java/org/xbill/DNS/Serial.java
ibauersachs: test_compare_boundary IMO asserts wrong values, arg2 is definitely not larger than arg1:

	long arg1 = 0xFFFFFFFFL;
	long arg2 = 0;
	int ret = Serial.compare( arg1, arg2 );
	assertEquals( -1, ret );

In master, we can simply replace the entire block with `return Long.compare(serial1, serial2);`, which reduces the method to sanity checking the inputs (which in turn is only necessary because of Java's horrific stupidity not having unsigned integers).
If this should be fixed in 2.1.x too, just compare the longs instead of calculating: `return s1 < s2? - 1 : (s1 == s2 ? 0 : 1);`

PR #48, opened 2019-05-30
Add Bouncy Castle test

kingle:
Basically stole the test from https://github.com/dnsjava/dnsjava/blob/6d50d8cc03e412642f28a0890c626f1283b8a86e/src/test/java/org/xbill/DNS/externalprovider/DNSSECWithLunaProviderTest.java#L1

but swapped with Bouncy Castle.  Is there still value in keeping `DNSSECWithLunaProviderTest` ?


pom.xml
```diff
@@ -179,6 +179,12 @@
     </build>
 
     <dependencies>
+        <dependency>
+            <groupId>org.bouncycastle</groupId>
+            <artifactId>bcprov-jdk15on</artifactId>
+            <version>1.60</version>
```
ibauersachs: Doesn't really matter for this use case, but 1.61 is available.

kingle: Bumped it up in latest commit

PR #49, opened 2019-05-31
fix maximum valid port number to 65535

abcdenis:


ibauersachs:
The maximum port number is now 65534. What are you trying to solve here?

PR #51, opened 2019-06-02
Adds ResolverConfigTest

kingle:
* Adding some tests to help with upcoming refactoring of ResolverConfig to allow SPI and also to sanity check the move of the properties files still works when testing on Windows.
* Changed many findX methods in ResolverConfig to return boolean on success/failure to aid in testing
* Minor change to add a package-private fromConstantString method to set origin, also to aid in testing
* Moved non-java related resources to appropriate `src/main/resources` and `src/test/resources`
* Removed spi-up-to-java8 profile -- unneeded now in the master branch


src/main/java/org/xbill/DNS/ResolverConfig.java
```diff
@@ -53,9 +56,7 @@
 		String OS = System.getProperty("os.name");
 		String vendor = System.getProperty("java.vendor");
 		if (OS.contains("Windows")) {
-			if (OS.contains("95") ||
-				OS.contains("98") ||
-				OS.contains("ME"))
+			if (Stream.of("95", "98", "ME").anyMatch(OS::contains))
```
kingle: Do we still want to support these ancient Windows versions?

ibauersachs: No

src/test/java/org/xbill/DNS/ResolverConfigTest.java
```diff
@@ -0,0 +1,88 @@
...
+	}
+
+	@Test
+	void findProperty_Success() {
+		String[] dnsServers = { "server1", "server2" };
+		// intentionally adding duplicate search entries for testing
+		String[] dnsSearch = { "dnsjava.org", "example.com",
+			"dnsjava.org" };
+		Name[] searchPath = Arrays.stream(dnsSearch)
+			.map(s -> Name.fromConstantString(s, Name.root))
```
ibauersachs: If this call is the only reason you added Name.fromConstantString(String, Name), why don't you simply put the trailing dot for the root to the string array?

src/test/java/org/xbill/DNS/ResolverConfigTest.java
```diff
@@ -0,0 +1,88 @@
...
+	}
+
+	@Test
+	@DisabledOnOs({OS.WINDOWS})
+	void findNT_NotWindows() {
+		assertFalse(ResolverConfig.getCurrentConfig().findWin());
+	}
+
+	@Test
+	@EnabledOnOs({OS.LINUX, OS.MAC, OS.AIX, OS.SOLARIS})
```
ibauersachs: Or @DisabledOnOs({OS.WINDOWS})? I doubt any test run will ever be executed on Android - if it's detected at all.

src/main/java/org/xbill/DNS/ResolverConfig.java
```diff
@@ -261,26 +259,32 @@ else if(line.startsWith("options")) {
...
 	configureFromLists(lserver, lsearch);
 	configureNdots(lndots);
+	return true;
 }
 
-private void
+boolean
 findUnix() {
-	findResolvConf("/etc/resolv.conf");
+	return findResolvConf("/etc/resolv.conf");
```
ibauersachs: Can we pass an InputStream to findResolvConf (and wrap the Files.newInputStream(Paths.get("...")) into an auto-close try)?

src/test/java/org/xbill/DNS/ResolverConfigTest.java
```diff
@@ -0,0 +1,88 @@
...
+	void findUnix() {
+		assertTrue(ResolverConfig.getCurrentConfig().findUnix());
+	}
+
+	@Test
+	void resolvConfLoaded() throws URISyntaxException {
+		assertTrue(ResolverConfig.getCurrentConfig()
+			.findResolvConf(Paths.get(getClass()
```
ibauersachs: ResolverConfigTest.class.getResourceAsStream("/...")

pom.xml
```diff
@@ -194,24 +194,6 @@
     </dependencies>
 
     <profiles>
-        <profile>
```
ibauersachs: The service descriptor still needs to be excluded from the JAR on Java 9+. But it can probably be moved to the no-spi-on-java9 profile with an exclusion instead of an inclusion if you move it (correctly so, I missed that).

src/test/java/org/xbill/DNS/ResolverConfigTest.java
kingle: Future tests may look at the inputs from other sources for search path (ipconfig output), and those sources usually don't have the trailing dot.  We could adjust the inputs, but I feel conflicted relying on massaging test resources to have trailing dots.  That being said, it's your call.  I can go either way.

src/main/java/org/xbill/DNS/ResolverConfig.java
```diff
@@ -53,12 +56,7 @@
 		String OS = System.getProperty("os.name");
 		String vendor = System.getProperty("java.vendor");
 		if (OS.contains("Windows")) {
-			if (OS.contains("95") ||
-				OS.contains("98") ||
-				OS.contains("ME"))
-				find95();
-			else
-				findNT();
+			findWin();
```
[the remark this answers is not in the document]
kingle: Yes

src/main/java/org/xbill/DNS/ResolverConfig.java
```diff
@@ -355,71 +359,62 @@ else if(line.startsWith("options")) {
...
-private void
-findNT() {
+boolean
+findWin() {
+	boolean found;
 	try {
 		Process p;
 		p = Runtime.getRuntime().exec("ipconfig /all");
-		findWin(p.getInputStream());
+		found = findWin(p.getInputStream());
```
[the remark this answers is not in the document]
kingle: netsh is still locale-dependent (on my Windows 10 with Spanish set):
```
C:\...>netsh int ipv4 show dnsservers
Configuración para la interfaz "Ethernet"
    Servidores DNS configurados a través de DHCP:  192.168.1.1
    Registrar con el sufijo:           Solo el principal
...
```

That said, it's probably a better source to parse since it eliminates most of the noise

.gitignore
```diff
@@ -4,4 +4,5 @@
 .idea/
 *.iml
 target/
+.DS_Store
```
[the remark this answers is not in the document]
kingle: yep, still exists even on macOS Mojave

pom.xml
kingle: Good call.  Excluded in that profile.

src/main/java/org/xbill/DNS/ResolverConfig.java
kingle: Yes!  Done in latest commit.

src/test/java/org/xbill/DNS/ResolverConfigTest.java
kingle: Agreed.  Changed in latest

src/main/java/org/xbill/DNS/ResolverConfig.java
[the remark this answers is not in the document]
ibauersachs: It lacks the search domains though, not sure atm. if there's another netsh call for those.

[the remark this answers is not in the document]
kingle: I still think we should consider the registry option.  What's wrong with looking at:
`regedit /e TEMPFILE "HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Services\Tcpip\Parameters"`
and parse those?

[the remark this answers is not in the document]
ibauersachs: You'll get nameservers of NICs with a disconnected link. Example: laptop with a wired and WiFi NIC. WiFi is connected at home to your router and the DHCP server assigns an IP and DNS addresses. Later you go to work, connect via cable (e.g. docking station), WiFi is disconnected. Your regedit call (and what Java does internally with probing the registry and calling Windows 95 (!) networking APIs) will get the inaccessible nameservers of your home WiFi.
Or the other way around (office nameservers still on the wired NIC, now only a WiFi connection). Querying the registry for this info is just _completely_ wrong. See also [JDK-7006496](https://bugs.openjdk.java.net/browse/JDK-7006496).

What I'd be fine with though is adding an _optional_ (i.e. scope=provided) dependency to JNA, probing at runtime if it's available and then doing proper calls to Iphlpapi. But that API isn't easy.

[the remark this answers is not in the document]
kingle: name servers sure, but what about search suffixes?  Regardless of whether you disconnect and reconnect at different places, do the DNS Search Suffixes change?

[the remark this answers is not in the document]
ibauersachs: Yes, they do change, at least their priority. At work I have `ds.example.com`, at home my cable router hands out `home`. Fritz Boxes (popular routers/modems in Europe and especially Germany) assign `fritz.box`.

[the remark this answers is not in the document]
kingle: Interesting.  Can we table this for a different PR/issue?  How we go about adding new functionality to better handle Windows resolution specifically could go on for awhile and we probably should discuss it in a separate issue.  I definitely don't want to tackle native code additions here.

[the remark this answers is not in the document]
ibauersachs: Sure, this was just about the question if netsh is also locale dependent. It is, so, bummer.

src/test/java/org/xbill/DNS/ResolverConfigTest.java
ibauersachs: The overload is anyway not necessary, this is equivalent: `Name[] searchPath = Arrays.stream(dnsSearch).map(s -> Name.fromConstantString(s + "."))`

src/main/java/org/xbill/DNS/ResolverConfig.java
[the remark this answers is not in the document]
kingle: Logged issue #55 to track the work to call Iphlpapi

src/test/resources/test_loaded_resolv.conf
```diff
@@ -0,0 +1,5 @@
+# comment
+nameserver 192.168.1.1
+domain domain.com
+search example.com dnsjava.org
+options ndots:5
```
ibauersachs: Interesting. I wasn't aware there were this many options for resolv.conf. I wonder if we want to support some of them in the future (not for this PR, just a remark).

kingle: Logged issue #57

src/test/java/org/xbill/DNS/ResolverConfigTest.java
kingle: Updated and reverted the Name addition

src/main/java/org/xbill/DNS/ResolverConfig.java
```diff
@@ -30,9 +31,11 @@
  * @author <a href="mailto:yannick@meudal.net">Yannick Meudal</a>
  * @author <a href="mailto:arnt@gulbrandsen.priv.no">Arnt Gulbrandsen</a>
  */
-
 public class ResolverConfig {
 
+	static final String DNS_SERVER_PROP = "dns.server";
```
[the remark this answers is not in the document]
kingle: Made public, dropped indent (when is the mass test formatting happening?)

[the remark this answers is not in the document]
ibauersachs: Not sure yet, if no pull requests are open. And only on test sources anyway. I haven't decided to apply it on the main sources yet since they're not so messy.

PR #56, opened 2019-06-03
Lists for RRset

ibauersachs:
Do you agree with the new API on RRset? This is the first real breaking change. I thought about adding an overload or making rrs() generic if the set is expected to contain only one type.

### DIFF
--- a/src/main/java/org/xbill/DNS/Message.java
+++ b/src/main/java/org/xbill/DNS/Message.java
@@ -365,7 +365,7 @@ public RRset []
 getSectionRRsets(int section) {
 	if (sections[section] == null)
 		return emptyRRsetArray;
-	List<RRset> sets = new LinkedList<>();
+	List<RRset<Record>> sets = new LinkedList<>();
 	Record [] recs = getSectionArray(section);
 	Set<Name> hash = new HashSet<>();
 	for (Record rec : recs) {
@@ -373,7 +373,7 @@ getSectionRRsets(int section) {
 		boolean newset = true;
 		if (hash.contains(name)) {
 			for (int j = sets.size() - 1; j >= 0; j--) {
-				RRset set = sets.get(j);
+				RRset<Record> set = sets.get(j);
 				if (set.getType() == rec.getRRsetType() &&
 					set.getDClass() == rec.getDClass() &&
 					set.getName().equals(name))
@@ -385,7 +385,7 @@ getSectionRRsets(int section) {
 			}
 		}
 		if (newset) {
-			RRset set = new RRset(rec);
+			RRset<Record> set = new RRset<>(rec);
 			sets.add(set);
 			hash.add(name);
 		}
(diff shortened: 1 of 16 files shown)


bwelling:
Is there some reason to return a List instead of an Iterator?  That seems like extra copying.

RRSets always contain records of only one type (not including signatures, which are separate).

ibauersachs:
@bwelling Right, seems it was a little late when I wrote that. The type-bitmap in RRSIGs must have confused me. The lists aren't copied again, that is already being done in the existing private rrs(boolean, boolean) to split data from sigs. There could be two separate internal Lists to make this unnecessary in the non-cycling case.
Do you remember the intention behind the cycling? I understand why that needs to be done in general, but not why this shouldn't be application specific.

src/main/java/org/xbill/DNS/RRset.java
```diff
@@ -150,32 +151,35 @@
...
 }
 
 /**
  * Returns an Iterator listing all (data) records.
  * @param cycle If true, cycle through the records so that each Iterator will
  * start with a different record.
  */
-public synchronized Iterator<Record>
+public synchronized List<Record>
 rrs(boolean cycle) {
```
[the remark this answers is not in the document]
ibauersachs: I thought about keep the existing methods as deprecated, but really liked their short and concise name. While still standard Java, getRrs and getSigs is somewhat cumbersome. Not really an argument, I know.
If consumers don't want to refactor their existing iterator code to for-each's, a simple regex - `s/\.rrs\(\)/\.rrs\(\)\.iterator\(\)/` - will do. So there's not much refactoring involved.

Not sure yet if it's possible to make the entire RRset class generic. But the casting that I need to do everywhere in dnssecjava is reason enough to try.

bwelling:
I'm not sure why the rrs and sigs lists are separate.  It's possible that it was a misguided optimization, but it would definitely make sense to split them.

The optional cycling is there because iterating through records in a nondeterministic order is often useful, so it seemed reasonable to make it part of the RRset iteration.  If I was designing the API from scratch, I'm not sure if it would still make sense or not (it should be somewhere, but that might njot be the right place), and I definitely wouldn't implement the existing cyclic behavior that requires a write lock to maintain the position counter.

src/main/java/org/xbill/DNS/Message.java
```diff
@@ -364,15 +365,15 @@
 getSectionRRsets(int section) {
 	if (sections[section] == null)
 		return emptyRRsetArray;
-	List<RRset> sets = new LinkedList<>();
+	List<RRset<Record>> sets = new LinkedList<>();
```
kingle: Is there a particular reason to use a LinkedList here instead of ArrayList?

ibauersachs: I don't think so. It's been like that for 17 years...

PR #60, opened 2019-06-09
Add more tests for zone files

kingle:


ibauersachs:
Not that I mind more test, but aren't you basically testing `Record.fromString` instead of actually testing `Master`?

kingle:
I don't think I follow you.  The tests go off a zone file with a lot of untested records that are loaded by Master.  What are you proposing?

ibauersachs:
> I don't think I follow you. The tests go off a zone file with a lot of untested records that are loaded by Master. What are you proposing?

All that Master does (should do...) is delegating record parsing to the actual record type. I'd test the so-far untested record types in their respective test classes (or create one if it doesn't exist yet). And test Master for what it should do with special instructions only. E.g. this just asserts that ISDNRecord parses the what it gets from Master:

```
			assertEquals(Type.ISDN, rr.getType());
			assertEquals("150862028003217", ((ISDNRecord) rr).getAddress());
			assertEquals("004", ((ISDNRecord) rr).getSubAddress());
```

While e.g. `includeDirective` and `expandGenerated` are great and really test Master.

PR #67, opened 2019-07-09
Add support for disabling $INCLUDE

jdreed:
- Add support for disabling $INCLUDE when parsing master files
  - Optionally throw an exception (for bubbling up) when $INCLUDE processing is disabled, and the token is encountered

### DIFF
--- a/src/main/java/org/xbill/DNS/Master.java
+++ b/src/main/java/org/xbill/DNS/Master.java
@@ -408,8 +408,7 @@ nextRecord() throws IOException {
  */
 public void
 disableIncludes() {
-	noExpandIncludes = true;
-	includeThrowsException = false;
+	disableIncludes(false);
 }
 
 /**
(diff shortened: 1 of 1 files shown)


src/main/java/org/xbill/DNS/Master.java
```diff
@@ -391,6 +399,34 @@ else if ((token.value).charAt(0) == '$') {
...
+/**
+ * Disable processing of $INCLUDE directives.
+ * When disabled, $INCUDE statements will not be processed.
+ * Depending on the contents of the file that would have been included,
+ * this may cause the zone to be invalid.
+ * (e.g. if there is no SOA or NS at the apex)
+ */
+public void
+disableIncludes() {
+	noExpandIncludes = true;
```
ibauersachs: Please call the overload instead of directly assigning variables: `disableIncludes(false)`

jdreed: Updated with d9944b2.    What is this project's preferred workflow?   rebase and re-push to my branch?   Or keep both commits?

ibauersachs: Doesn't really matter to me. I often squash PRs during merge if there's no relevant history.

PR #68, opened 2019-07-13
Move sun.net Resolver down

kingle:
Moving the sun.net.dns.ResolverConfiguration down the ordering

### DIFF
--- a/src/main/java/org/xbill/DNS/ResolverConfig.java
+++ b/src/main/java/org/xbill/DNS/ResolverConfig.java
@@ -19,9 +19,9 @@ import java.util.stream.Stream;
  *       are checked.  The servers can either be IP addresses or hostnames
  *       (which are resolved using Java's built in DNS support).
  *   <LI>On Unix, /etc/resolv.conf is parsed.
+ *   <LI>The sun.net.dns.ResolverConfiguration class is queried.
  *   <LI>On Windows, ipconfig is called and its output parsed.  This
  *       may fail for non-English versions on Windows.
- *   <LI>The sun.net.dns.ResolverConfiguration class is queried.
  *   <LI>"localhost" is used as the nameserver, and the search path is empty.
  * </UL>
  *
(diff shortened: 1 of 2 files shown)


kingle:
The driving force is that on Unix-like systems, the preference of order would be to use resolv.conf before trying the sun implementation, for ndots.  Your point is taken about Windows.  If this is changed to try the sun resolver on Windows first and fallback to ipconfig, would that suffice?

As far as the SPI work, I was hoping the OpenJDK itself would progress further on the (almost identical) interface they initially suggested: https://bugs.openjdk.java.net/browse/JDK-8211216

ibauersachs:
`resolv.conf` for UNIX-like is fine with me. Forget about hoping the JDK will act on this anytime soon, and we need a solution for JDK 11 anyway since this is what Ubuntu 18.04 and Debian Buster are shipping in their stable/LTS releases.

src/main/java/org/xbill/DNS/ResolverConfig.java
```diff
@@ -52,21 +52,25 @@
 ResolverConfig() {
 	if (findProperty())
 		return;
-	if (findSunJVM())
-		return;
+	boolean attemptedSun = false;
 	if (servers == null || searchlist == null) {
```
ibauersachs: This check is now useless - findProperty is ony true when either servers or searchList is null

src/main/java/org/xbill/DNS/ResolverConfig.java
```diff
@@ -52,21 +52,25 @@
...
-			findNetware();
-		} else if (vendor.contains("Android")) {
-			findAndroid();
-		} else {
-			findUnix();
-		}
+			if (findSunJVM() || findWin())
+				return;
+			else
+				attemptedSun = true;
```
ibauersachs: I don't think `attemptedSun` makes sense like this. It's only used once, so the code would be simpler to read if you just add

    else
        findSunJVM();

as the general fallback if none of the OS checks match.

PR #76, opened 2019-09-24
Fix algorithm check to not assume software key instances to work with PKCS#11

netmackan:
DNSSEC signing does not currently work with PKCS#11 keys as the checkAlgorithm() method assumes the key instance is a software key and not an unextractable key in an HSM.

When providing a private key instance from SunPKCS11 such as a P11PrivateKey the following exception is thrown:
> org.xbill.DNS.DNSSEC$IncompatibleKeyException: incompatible keys
>	at deployment.signserver.ear//org.xbill.DNS.DNSSEC.checkAlgorithm(DNSSEC.java:1071)
>	at deployment.signserver.ear//org.xbill.DNS.DNSSEC.sign(DNSSEC.java:1129)

This patch suggests changing from checking the key type using instanceof, as one can not know which instances will be provided and corresponds to which key type, to instead use the Key.getAlgorithm()  method. https://docs.oracle.com/javase/9/docs/api/java/security/Key.html#getAlgorithm--

This way signing also works with PKCS#11.


netmackan:
The build failure is most likely due to that this DNSSEC.java file was not following the styleguide even before this change, right?
Let me know if anything needs to be changed.

ibauersachs:
The build fails because the now unused imports `java.security.interfaces.EC/RSA/DSAPrivateKey` weren't removed. Do you mind fixing that?

netmackan:
Thanks that was it! Fixed now.

PR #78, opened 2019-10-14
addition of EDNS Cookie and TCP Keepalive Options

demonti:
I have implemented EDNS options for DNS Cookies (RFC 7873) and for TCP Keepalive (RFC 7828). When adding the BADCOOKIE RCODE to the Rcode class, I discovered that for TKEY, two RCODEs were missing and and another was added to the tsigrcodes instance, which is IMHO wrong. I moved the latter to the regular rcodes instance. As this is fully included into the tsigrcodes instance, it should not break any existing code. Finally, I fixed one test class, as the RCODE 20 is no longer a "reserved" value, but actually used.

It would be nice if you could eventually merge my additions and changes into the trunk. In case of any questions or problems, feel free to contact me.

### DIFF
--- a/src/main/java/org/xbill/DNS/CookieOption.java
+++ b/src/main/java/org/xbill/DNS/CookieOption.java
@@ -1,6 +1,7 @@
 package org.xbill.DNS;
 
 import java.io.IOException;
+import java.util.Optional;
 import org.xbill.DNS.utils.base16;
 
 /**
@@ -14,8 +15,8 @@ public class CookieOption extends EDNSOption {
   /** client cookie */
   private byte[] clientCookie;
 
-  /** server cookie, may be {@code null} */
-  private byte[] serverCookie;
+  /** server cookie */
+  private Optional<byte[]> serverCookie;
 
   /** Default constructor for constructing instance from binary representation. */
   CookieOption() {
@@ -28,25 +29,27 @@ public class CookieOption extends EDNSOption {
    * @param clientCookie the client cookie, which must consist of eight bytes
    */
   public CookieOption(byte[] clientCookie) {
-    this(clientCookie, null);
+    this(clientCookie, Optional.empty());
   }
 
   /**
    * Constructor.
    *
    * @param clientCookie the client cookie, which must consist of eight bytes
-   * @param serverCookie the server cookie, which either must be {@code null} or must consist of 8
-   *     to 32 bytes
+   * @param serverCookie the server cookie, which must consist of 8 to 32 bytes if present
    */
-  public CookieOption(byte[] clientCookie, byte[] serverCookie) {
+  public CookieOption(byte[] clientCookie, Optional<byte[]> serverCookie) {
     this();
-    if (clientCookie == null) throw new IllegalArgumentException("cookie must not be null");
+    if (clientCookie == null) throw new IllegalArgumentException("client cookie must not be null");
     if (clientCookie.length != 8)
-      throw new IllegalArgumentException("cookie must consist of eight bytes");
+      throw new IllegalArgumentException("client cookie must consist of eight bytes");
     this.clientCookie = clientCookie;
 
-    if (serverCookie != null && (serverCookie.length < 8 || serverCookie.length > 32))
-      throw new IllegalArgumentException("cookie must consist of 8 to 32 bytes");
+    if (serverCookie.isPresent()) {
+      int length = serverCookie.get().length;
+      if (length < 8 || length > 32)
+        throw new IllegalArgumentException("server cookie must consist of 8 to 32 bytes");
+    }
     this.serverCookie = serverCookie;
   }
 
@@ -62,9 +65,9 @@ public class CookieOption extends EDNSOption {
   /**
    * Returns the server cookie.
    *
-   * @return the server cookie, may be {@code null}
+   * @return the server cookie
    */
-  public byte[] getServerCookie() {
+  public Optional<byte[]> getServerCookie() {
     return serverCookie;
   }
 
@@ -82,7 +85,9 @@ public class CookieOption extends EDNSOption {
     if (length > 8) {
       if (length < 16 || length > 40)
         throw new WireParseException("invalid length of server cookie");
-      serverCookie = in.readByteArray();
+      serverCookie = Optional.of(in.readByteArray());
+    } else {
+      serverCookie = Optional.empty();
     }
   }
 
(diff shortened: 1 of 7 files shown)


demonti:
I have to admit that I am lost with git and github, this is not my world. Do I need to resubmit the merge request after the reformatting? Shall I squash all changes beforehand?

diegosalvi:
Hi @demonti, no you don't have to resubmit any merge/pull request. It already show in this thread and it will be ported into master if the whole PR will be accepted and merged.
The squash can be done from repository admin when merging the PR with the master.

src/main/java/org/xbill/DNS/CookieOption.java
```diff
@@ -0,0 +1,111 @@
...
+   *
+   * @param clientCookie the client cookie, which must consist of eight bytes
+   * @param serverCookie the server cookie, which either must be {@code null} or must consist of 8
+   *     to 32 bytes
+   */
+  public CookieOption(byte[] clientCookie, byte[] serverCookie) {
+    this();
+    if (clientCookie == null) throw new IllegalArgumentException("cookie must not be null");
+    if (clientCookie.length != 8)
+      throw new IllegalArgumentException("cookie must consist of eight bytes");
```
ibauersachs: `"client cookie...`

src/main/java/org/xbill/DNS/CookieOption.java
```diff
@@ -0,0 +1,111 @@
...
+    this.clientCookie = clientCookie;
+
+    if (serverCookie != null && (serverCookie.length < 8 || serverCookie.length > 32))
+      throw new IllegalArgumentException("cookie must consist of 8 to 32 bytes");
```
ibauersachs: `"server cookie...`

src/main/java/org/xbill/DNS/CookieOption.java
```diff
@@ -0,0 +1,111 @@
...
+  public byte[] getClientCookie() {
+    return clientCookie;
+  }
+
+  /**
+   * Returns the server cookie.
+   *
+   * @return the server cookie, may be {@code null}
+   */
+  public byte[] getServerCookie() {
```
ibauersachs: Please use `java.util.Optional`, i.e. `public java.util.Optional<byte[]> getServerCookie()`

src/main/java/org/xbill/DNS/TcpKeepaliveOption.java
```diff
@@ -0,0 +1,100 @@
...
+/**
+ * TCP Keepalive EDNS0 Option, as defined in https://tools.ietf.org/html/rfc7828
+ *
+ * @see OPTRecord
+ * @author Klaus Malorny
+ */
+public class TcpKeepaliveOption extends EDNSOption {
+
+  /** flag whether the timeout has been provided */
+  private boolean hasTimeout;
```
ibauersachs: Isn't this implicitly given if `timeout` is not zero?

src/main/java/org/xbill/DNS/TcpKeepaliveOption.java
```diff
@@ -0,0 +1,100 @@
...
+
+  /** the timeout in 100ms units */
+  private int timeout;
```
ibauersachs: Please use `java.time.Duration` and only encode to 100ms when needed (i.e. in `optionToWire`).

demonti:
@ibauersachs 
* unit tests: I can do this; I omitted them because there were no unit tests for the already existing EDNS options neither
* integration into SimpleResolver: I am using the new options only server-side and I am not very familiar with the resolver code contained in the library. I therefore would prefer to keep the SimpleResolver untouched
* Optional/Duration: I am still reluctant of using high-level constructs in low-level, performance critical context (where I use dnsjava) because of their resource toll. Also, Optional does not seem to be very popular, even among Oracle JDK API designers. In addition, I am against a mixed style library. Therefore I chose to continue the "traditional" style Brian Wellington used. If you intend to convert the whole API to Optional/Duration/Instant and such, this would be fine to me, however.
* regarding TCP keepalive, timeout value: yes, zero is a valid value, indicating that the client shall not further use the TCP connection. Alternatively to the flag or Optional, a negative value would be possible, but this is not my preference.

src/main/java/org/xbill/DNS/CookieOption.java
demonti: Sorry, I do not understand; what is exactly the problem with it (and "server cookie", too)?

ibauersachs: Please add `client` and `server` to the exception messages, otherwise it isn't clear to which parameter the exception refers to.

ibauersachs:
> * unit tests: I can do this; I omitted them because there were no unit tests for the already existing EDNS options neither

That's not really an argument to omit it for new code ;-)

> * integration into SimpleResolver: I am using the new options only server-side and I am not very familiar with the resolver code contained in the library. I therefore would prefer to keep the SimpleResolver untouched

Okay.

> * Optional/Duration: I am still reluctant of using high-level constructs in low-level, performance critical context (where I use dnsjava) because of their resource toll. Also, Optional does not seem to be very popular, even among Oracle JDK API designers. In addition, I am against a mixed style library. Therefore I chose to continue the "traditional" style Brian Wellington used. If you intend to convert the whole API to Optional/Duration/Instant and such, this would be fine to me, however.

I've already converted the DNSSEC API from `Date` to `Instant`, so using `Duration` is IMO a must. 

Using Optional is a much cleaner way of describing that the server cookie "can be null", but I guess you can talk me out of that.

> * regarding TCP keepalive, timeout value: yes, zero is a valid value, indicating that the client shall not further use the TCP connection. Alternatively to the flag or Optional, a negative value would be possible, but this is not my preference.

Okay, keep the boolean flag.

demonti:
> > * unit tests: I can do this; I omitted them because there were no unit tests for the already existing EDNS options neither
> 
> That's not really an argument to omit it for new code ;-)
> 
It was an attempt ;-)

[…]
> > * regarding TCP keepalive, timeout value: yes, zero is a valid value, indicating that the client shall not further use the TCP connection. Alternatively to the flag or Optional, a negative value would be possible, but this is not my preference.
> 
> Okay, keep the boolean flag.

If you want to use Optional anyway, I can use an OptionalInt for that. This would be consistent.

src/main/java/org/xbill/DNS/CookieOption.java
demonti: Ok, that makes sense. I didn't notice that.

demonti:
Just kindly want to ask whether there are any issues I have missed?

ibauersachs:
Sorry, this slipped.

I was wondering if the use of Optional for the serverCookie is appropriate. Optional is only meant for the API, not fields and static code analyzer rant about that.
Same actually for the timeout field, but there I'm even more tempted to just ignore it as having an additional field like in your first version isn't really better.

Issue #80, opened 2019-10-31
Resolver on non-default port


avmusa:
`org.xbill.DNS.SimpleResolver` and `org.xbill.DNS.ExtendedResolver` work only on a default port (53).
Whilst custom port can be set with `org.xbill.DNS.SimpleResolver.setPort(int)`, it is not supported in the `org.xbill.DNS.ExtendedResolver`.

When I pass `-Ddns.server=1.2.3.4:5300` it's failing because `1.2.3.4:5300` is used for the hostname with default port (53).

bwelling:
Both `SimpleResolver` and `ExtendedResolver` support alternate ports; `setPort()` is part of the `Resolver` interface, and is implemented by both.

The `dns.server` command line options do not support alternate ports, never have, and are not documented as such.  They were intended as a way to manually override the name servers typically provided by the operating system, which do not include alternate ports.

Even if they did support alternate ports, it would not be using the `1.2.3.4:5300` syntax, as that would not support IPv6 (an IPv6 address with a port would be ambiguous).

ibauersachs:
IPv6 addresses would need to be specified as e.g. `[2001:db8::1]:53`. The entire resolver configuration will be overhauled/rewritten for v3, see #8. Not sure yet if I'll do another 2.x release and if this would make it there.

Issue #57, opened 2019-06-04
Support more resolv.conf options/overrides


kingle:
According to the resolv.conf [man page](http://man7.org/linux/man-pages/man5/resolv.conf.5.html), there are a lot of options.  We should consider if we want to grab additional options, aside from the ndots we grab currently.

In addition, we should honor the environment variables overrides in the ResolverConfig:
```
       The search keyword of a system's resolv.conf file can be overridden
       on a per-process basis by setting the environment variable
       LOCALDOMAIN to a space-separated list of search domains.

       The options keyword of a system's resolv.conf file can be amended on
       a per-process basis by setting the environment variable RES_OPTIONS
       to a space-separated list of resolver options as explained above
       under options.
```

TimEvens:
+1 for RES_OPTIONS. 

We are running into an issue with ResolverConfiguration [ResolverConfig Line 55](https://github.com/dnsjava/dnsjava/blob/master/src/main/java/org/xbill/DNS/ResolverConfig.java#L55) calls ```findSunJVM()``` and if true, returns without calling ```findUnix()```  to load ```/etc/resolv.conf```.   While this works fine for **nameservers** and **search domains** it doesn't work for **ndots**.  Apparently ```findSunJVM()``` does not load options and therefore does not set the correct **ndots** value as configured.   ```InetAddress.getByName()``` works fine when loading ndots.

Issue #55, opened 2019-06-03
Use iphlpapi for Windows resolution


kingle:
ResolverConfig currently parses ipconfig to grab DNS information as a fallback.  With the sun ResolverConfiguration implementation possibly going away in the future, we should consider additional options.  The most promising one is to use the iphlpapi library if on Windows using the (optional scoped) JNA dependency.  Call `GetNetworkParams`:

C++ example code here:
https://docs.microsoft.com/en-us/windows/desktop/api/iphlpapi/nf-iphlpapi-getnetworkparams

kingle:
Wow.  Looks like JNA basically did this for us: https://github.com/java-native-access/jna/blob/cc1acdac02e4d0dda93ba01bbe3a3435b8933dab/contrib/platform/test/com/sun/jna/platform/win32/IPHlpAPITest.java#L114

ibauersachs:
I was looking at a too old JNA version (some 4.x) yesterday and missed that it's now there. Awesome, something less to worry about.

kingle:
One caveat, when I did the testing, it looks like the `GetNetworkParams` only works with IPv4 addresses and doesn't return the IPv6 address like my ipconfig outputs.  There may still need to be some work to make it a better "try-this-before-ipconfig" option on Windows.  We may want to look at the `GetAdaptersAddresses` https://docs.microsoft.com/en-us/windows/desktop/api/iphlpapi/nf-iphlpapi-getadaptersaddresses

ibauersachs:
Just another example of where Java's built-in mechanism for nameserver lookup fails (via Netty/JNDA):
- At home, connected via WiFi and Ethernet
- Both NICs have the providers nameservers assigned via DHCP
- The nameservers are public IPs, but are only accessible from within my ISPs network
- A VPN connection to my office assigns the companys nameservers
- The VPN is the default gateway (with metric=0), so everything goes via the VPN

Netty tries to query the ISPs nameservers and fails since the query reaches them via the public IP of my company. Yuck. Interestingly though, InetAddress.getByName("...") works - `sun.net.dns.ResolverConfiguration.open().nameservers()` contains seven nameservers, including the ones assigned via VPN (IPv4 only, the VPN is dual-stack and assigns nameservers).

ibauersachs:
See branch `windows-dns-search` for a demo of retrieving all DNS servers (IPv4/IPv6) via JNA. Needs cleanup and ordering of interfaces by metric.

PR #82, opened 2019-11-04
Refactor nameserver lookup to an SPI interface

### LINKED ISSUE #6: Android O support
Since Android O removed net.dns1/2/3/4, now the dns look up dose not work on Android O devices, could you introduce Android O support?

ibauersachs:
Closes #6, #8, #55, #57, #68, #80


src/main/java/org/xbill/DNS/config/AndroidResolverConfigProvider.java
```diff
@@ -0,0 +1,115 @@
...
+import lombok.extern.slf4j.Slf4j;
+import org.xbill.DNS.ResolverConfig;
+import org.xbill.DNS.SimpleResolver;
+
+/**
+ * Resolver config provider for Android. Contrary to all other providers, this provider needs a
+ * context to operate on which must be set by calling {@link #setContext(Object)}.
+ *
+ * <p>If you are developing for Android, consider implementing your own {@link
+ * ResolverConfigProvider} that listens to network callbacks and properly refreshes on link changes.
```
camhart: Can you define "link changes" further?  Does that just mean network changes?

ibauersachs: Yes, I'm not developing for Android, so I can only assume things from reading the docs. `ConnectivityManager.registerNetworkCallback` and `registerDefaultNetworkCallback` seem to be interesting APIs.

aaronjwood:
Will this be released soon? I'm another victim of Android 8+ changes but it looks like this addresses #6

ibauersachs:
I don't have an ETA yet. It would help if you could test the current master branch and provide feedback. But as mentioned in the Javadoc of the Android provider, it would be best if you properly handle Androids' network callbacks in your own application.

aaronjwood:
Ok thanks.

PR #85, opened 2019-12-03
changed to java.io so it works in old android

gferraro:
This makes it work in all android versions as java.nio only exists in android 24 and above.

### DIFF
--- a/src/main/java/org/xbill/DNS/config/ResolvConfResolverConfigProvider.java
+++ b/src/main/java/org/xbill/DNS/config/ResolvConfResolverConfigProvider.java
@@ -2,12 +2,12 @@
 package org.xbill.DNS.config;
 
 import java.io.BufferedReader;
+import java.io.File;
+import java.io.FileInputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.net.InetSocketAddress;
-import java.io.File;
-import java.io.FileInputStream;
 import java.util.StringTokenizer;
 import org.xbill.DNS.SimpleResolver;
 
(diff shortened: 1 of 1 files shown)


ibauersachs:
Android doesn't have a resolv.conf anyway, so it would be better to disable this provider entirely.

dnsjava uses NIO for the resolver since 2005. How is this working if there's no NIO?

gferraro:
> Android doesn't have a resolv.conf anyway, so it would be better to disable this provider entirely.
> 
> dnsjava uses NIO for the resolver since 2005. How is this working if there's no NIO?

It has been using the 2005 version. Will look into disablnig this provider

gferraro:
> > Android doesn't have a resolv.conf anyway, so it would be better to disable this provider entirely.
> > dnsjava uses NIO for the resolver since 2005. How is this working if there's no NIO?
> 
> It has been using the 2005 version. Will look into disablnig this provider

gferraro:
> Android doesn't have a resolv.conf anyway, so it would be better to disable this provider entirely.
> 
> dnsjava uses NIO for the resolver since 2005. How is this working if there's no NIO?

How would one go about detecting if it was running on android and then disabling that provider? Is there a big problem in just using java.io instead?

ibauersachs:
> > Android doesn't have a resolv.conf anyway, so it would be better to disable this provider entirely.
> > dnsjava uses NIO for the resolver since 2005. How is this working if there's no NIO?
> 
> How would one go about detecting if it was running on android and then disabling that provider? 

Is returning https://github.com/dnsjava/dnsjava/blob/master/src/main/java/org/xbill/DNS/config/ResolvConfResolverConfigProvider.java#L126 false here not sufficient?

> Is there a big problem in just using java.io instead?

No, but I'm somewhat tired of doing special, in this case legacy, stuff for Android.

PR #94, opened 2020-03-09
custom record support

demonti:
Hi Ingo,
as discussed via e-mail, I hereby submit a patch that allows the use of own record implementations for "private use" record types. As some kind of a by-product I extended the idea somewhat so that also included record types can be overridden (although I have no use for it at the moment). As always, if there is anything that needs to be fixed from your perspective, please let me know.

### DIFF
--- a/src/main/java/org/xbill/DNS/Mnemonic.java
+++ b/src/main/java/org/xbill/DNS/Mnemonic.java
@@ -101,6 +101,17 @@ class Mnemonic {
   }
 
   /**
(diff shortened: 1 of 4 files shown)


demonti:
hmm, I did not notice that you have made the change on your own yesterday. So the pull request is obsolete now.

ibauersachs:
Sorry about the duplicate effort. I didn't realize you were already working on it since you mentioned you might need some time.
This contains some useful changes I didn't think of in my hasty changes, so I'll go over it again.

PR #96, opened 2020-03-16
Only select for tcp write when there is something to write

### LINKED ISSUE #95: Lookups stop working after a query that switches to TCP
We've recently updated our dependency on `dnsjava` from `2.1.8` to `3.0.1` in https://github.com/spotify/dns-java/pull/38. But when testing this out with some services, we started seeing timeouts on lookups.

After some debugging, I managed to isolate a scenario that causes this to happen.  In our case it's caused by SRV queries on names that contain many records. We see that it first gets a truncated response over UDP and then switches to TCP. This query works fine in itself, but all subsequent queries hang for some time and hit the timeout.

It's hard to reproduce if you don't have a set of SRV records available that would trigger a switch to TCP. But here's the code that reproduces it consistently for us:

```java
  public static void main(String[] args) throws TextParseException, InterruptedException {
    final String name = "_redacted1._tcp.redacted.net"; // this name contains 28 SRV records
    final String name2 = "_redacted2._tcp.redacted.net"; // this name contains i SRV records

    final Lookup lookup = new Lookup(name, Type.SRV, DClass.IN);
    final Record[] records = lookup.run();
    if (records != null) {
      System.out.println("records.length = " + records.length);
    } else {
      System.out.println("records is null");
    }

    final Lookup lookup2 = new Lookup(name2, Type.SRV, DClass.IN);
    final Record[] records2 = lookup2.run();
    if (records2 != null) {
      System.out.println("records2.length = " + records2.length);
    } else {
      System.out.println("records2 is null");
    }
  }
```

The output is
```
2020-03-16 19:53:38:229 +0100 [main] DEBUG org.xbill.DNS.config.ResolvConfResolverConfigProvider - Added spotify.net. to search paths
2020-03-16 19:53:38:230 +0100 [main] DEBUG org.xbill.DNS.config.ResolvConfResolverConfigProvider - Added /1.1.1.1:53 to nameservers
2020-03-16 19:53:38:289 +0100 [main] DEBUG org.xbill.DNS.Lookup - lookup _redacted1._tcp.redacted.net. SRV, cache answer: unknown
2020-03-16 19:53:38:328 +0100 [main] DEBUG org.xbill.DNS.ExtendedResolver - Sending _redacted1._tcp.redacted.net./SRV, id=27665 to resolver 0 (SimpleResolver [/1.1.1.1:53]), attempt 1 of 3
2020-03-16 19:53:38:330 +0100 [main] DEBUG org.xbill.DNS.SimpleResolver - Sending _redacted1._tcp.redacted.net./SRV, id=27665 to udp/1.1.1.1:53
2020-03-16 19:53:38:332 +0100 [main] DEBUG org.xbill.DNS.Client - Starting dnsjava NIO selector thread
2020-03-16 19:53:38:497 +0100 [ForkJoinPool.commonPool-worker-3] DEBUG org.xbill.DNS.SimpleResolver - Got truncated response for id 27665, retrying via TCP
2020-03-16 19:53:38:497 +0100 [ForkJoinPool.commonPool-worker-3] DEBUG org.xbill.DNS.SimpleResolver - Sending _redacted1._tcp.redacted.net./SRV, id=27665 to tcp/1.1.1.1:53
2020-03-16 19:53:38:792 +0100 [main] DEBUG org.xbill.DNS.Cache - caching successful for _redacted1._tcp.redacted.net.
2020-03-16 19:53:38:793 +0100 [main] DEBUG org.xbill.DNS.Lookup - Queried _redacted1._tcp.redacted.net. SRV: successful
2020-03-16 19:53:38:800 +0100 [main] DEBUG org.xbill.DNS.Lookup - lookup _redacted2._tcp.redacted.net. SRV, cache answer: unknown
2020-03-16 19:53:38:800 +0100 [main] DEBUG org.xbill.DNS.ExtendedResolver - Sending _redacted2._tcp.redacted.net./SRV, id=31420 to resolver 0 (SimpleResolver [/1.1.1.1:53]), attempt 1 of 3
2020-03-16 19:53:38:800 +0100 [main] DEBUG org.xbill.DNS.SimpleResolver - Sending _redacted2._tcp.redacted.net./SRV, id=31420 to udp/1.1.1.1:53
records.length = 28

// freezing for 10 seconds

2020-03-16 19:53:48:808 +0100 [main] DEBUG org.xbill.DNS.Lookup - Lookup failed using resolver ExtendedResolver of [SimpleResolver [/1.1.1.1:53]]
java.net.SocketTimeoutException
	at org.xbill.DNS.Resolver.send(Resolver.java:156)
	at org.xbill.DNS.Lookup.lookup(Lookup.java:493)
	at org.xbill.DNS.Lookup.resolve(Lookup.java:543)
	at org.xbill.DNS.Lookup.run(Lookup.java:561)
	at org.xbill.DNS.BasicUse.main(BasicUse.java:18)
2020-03-16 19:53:48:809 +0100 [main] DEBUG org.xbill.DNS.Lookup - lookup _redacted2._tcp.redacted.net.spotify.net. SRV, cache answer: unknown
2020-03-16 19:53:48:809 +0100 [main] DEBUG org.xbill.DNS.ExtendedResolver - Sending _redacted2._tcp.redacted.net.spotify.net./SRV, id=58929 to resolver 0 (SimpleResolver [/1.1.1.1:53]), attempt 1 of 3
2020-03-16 19:53:48:810 +0100 [main] DEBUG org.xbill.DNS.SimpleResolver - Sending _redacted2._tcp.redacted.net.spotify.net./SRV, id=58929 to udp/1.1.1.1:53
2020-03-16 19:53:58:813 +0100 [main] DEBUG org.xbill.DNS.Lookup - Lookup failed using resolver ExtendedResolver of [SimpleResolver [/1.1.1.1:53]]
java.net.SocketTimeoutException
	at org.xbill.DNS.Resolver.send(Resolver.java:156)
	at org.xbill.DNS.Lookup.lookup(Lookup.java:493)
	at org.xbill.DNS.Lookup.resolve(Lookup.java:543)
	at org.xbill.DNS.Lookup.run(Lookup.java:568)
	at org.xbill.DNS.BasicUse.main(BasicUse.java:18)
records2 is null
```

Debugging this a bit lead me to `org.xbill.DNS.Client`. After the TCP Transaction is done, the NIO selector thread goes into a busy spin where `selector.select(..)` always returns `1` signalling that the TCP channel is writable. But since there's nothing to write, the `processReadyKey` method on `ChannelState` just returns. I have not debugged further, but I noticed this since the above program started using one core at 100% (the selector thread). I'm guessing this is what is causing subsequent requests not to process and timeout.

rouzwawi:
Avoids putting the selector loop in full spin by removing the `OP_WRITE` selection key when there's nothing to write. I also removed the `registrationQueue` from the tcp client since we can just call connect immediately as we create the socket, and register the `OP_CONNECT` key there.

---
fixes #95

### DIFF
--- a/src/main/java/org/xbill/DNS/NioTcpClient.java
+++ b/src/main/java/org/xbill/DNS/NioTcpClient.java
@@ -6,6 +6,7 @@ import java.io.IOException;
 import java.net.InetSocketAddress;
 import java.net.SocketTimeoutException;
 import java.nio.ByteBuffer;
+import java.nio.channels.ClosedChannelException;
 import java.nio.channels.SelectionKey;
 import java.nio.channels.SocketChannel;
 import java.time.Duration;
@@ -23,6 +24,7 @@ import lombok.extern.slf4j.Slf4j;
 @Slf4j
 @UtilityClass
 final class NioTcpClient extends Client {
+  private static Queue<ChannelState> registrationQueue;
   private static Map<ChannelKey, ChannelState> channelMap;
   private static volatile boolean run;
 
@@ -32,9 +34,26 @@ final class NioTcpClient extends Client {
     }
 
     run = true;
+    registrationQueue = new ConcurrentLinkedQueue<>();
     channelMap = new ConcurrentHashMap<>();
+    addSelectorTimeoutTask(NioTcpClient::processPendingRegistrations);
     addSelectorTimeoutTask(NioTcpClient::checkTransactionTimeouts);
     start();
+  }
+
+  private static void processPendingRegistrations() {
+    while (!registrationQueue.isEmpty()) {
+      ChannelState state = registrationQueue.remove();
+      try {
+        if (!state.channel.isConnected()) {
+          state.channel.register(selector, SelectionKey.OP_CONNECT, state);
+        } else {
+          state.channel.keyFor(selector).interestOps(SelectionKey.OP_WRITE);
+        }
+      } catch (ClosedChannelException e) {
+        state.handleChannelException(e);
+      }
+    }
   }
 
   private static void checkTransactionTimeouts() {
@@ -54,6 +73,7 @@ final class NioTcpClient extends Client {
       return;
     }
 
+    registrationQueue.clear();
     EOFException closing = new EOFException("Client is closing");
     channelMap.forEach((key, state) -> state.handleTransactionException(closing));
     channelMap.clear();
@@ -145,8 +165,8 @@ final class NioTcpClient extends Client {
 
     private void processConnect(SelectionKey key) {
       try {
-        key.interestOps(SelectionKey.OP_WRITE);
         channel.finishConnect();
+        key.interestOps(SelectionKey.OP_WRITE);
       } catch (IOException e) {
         handleChannelException(e);
       }
@@ -206,7 +226,6 @@ final class NioTcpClient extends Client {
     }
 
     private void processWrite(SelectionKey key) {
-      key.interestOps(SelectionKey.OP_READ);
       for (Iterator<Transaction> it = pendingTransactions.iterator(); it.hasNext(); ) {
         Transaction t = it.next();
         try {
@@ -216,6 +235,7 @@ final class NioTcpClient extends Client {
           it.remove();
         }
       }
+      key.interestOps(SelectionKey.OP_READ);
     }
   }
 
@@ -248,10 +268,8 @@ final class NioTcpClient extends Client {
                     c.bind(local);
                   }
 
-                  final ChannelState channelState = new ChannelState(c);
-                  c.register(selector, SelectionKey.OP_CONNECT, channelState);
                   c.connect(remote);
-                  return channelState;
+                  return new ChannelState(c);
                 } catch (IOException e) {
                   f.completeExceptionally(e);
                   return null;
@@ -264,11 +282,7 @@ final class NioTcpClient extends Client {
             Type.string(query.getQuestion().getType()));
         Transaction t = new Transaction(query, data, endTime, channel.channel, f);
         channel.pendingTransactions.add(t);
-
-        // add read interest to the selection
-        final SelectionKey selectionKey = channel.channel.keyFor(selector);
-        selectionKey.interestOps(selectionKey.interestOps() | SelectionKey.OP_WRITE);
-
+        registrationQueue.add(channel);
         selector.wakeup();
       }
     } catch (IOException e) {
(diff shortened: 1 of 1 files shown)


rouzwawi:
@ibauersachs

ibauersachs:
Thanks for the investigation! I'll try to look at this asap. I assume with this changes your scenario is now working correctly?

src/main/java/org/xbill/DNS/NioTcpClient.java
```diff
@@ -265,8 +248,10 @@ private void processWrite() {
                     c.bind(local);
                   }
 
+                  final ChannelState channelState = new ChannelState(c);
+                  c.register(selector, SelectionKey.OP_CONNECT, channelState);
```
ibauersachs: AFAIR this won't work in a multithreaded scenario. A registration on the selector must be made on the selector thread or it can cause deadlocks.

rouzwawi: I guess you're referring to this in the [`Selector`](https://docs.oracle.com/javase/7/docs/api/java/nio/channels/Selector.html) docs:

> A selector's key and selected-key sets are not, in general, safe for use by multiple concurrent threads. If such a thread might modify one of these sets directly then access should be controlled by synchronizing on the set itself. The iterators returned by these sets' iterator methods are fail-fast: If the set is modified after the iterator is created, in any way except by invoking the iterator's own remove method, then a ConcurrentModificationException will be thrown.

Since [`SelectableChannel.register`](https://docs.oracle.com/javase/7/docs/api/java/nio/channels/SelectableChannel.html#register(java.nio.channels.Selector,%20int)) modifies the selectors key set, it might cause that ConcurrentModificationException to be thrown. But in general it seems that the register method itself can be called from multiple threads.

Anyhow, I think I will revert back to using the registration queue to avoid having to deal with concurrent modification exceptions on the selector thread.

ibauersachs: It's been a while, but I think I actually ran into deadlocks (not the ConcurrentModificationException) when registering from outside the selector thread.

rouzwawi: fixed

rouzwawi:
Any ETA on when this can be merged/released?

rouzwawi:
> I assume with this changes your scenario is now working correctly?

Yes I’ve tested a local snapshot build with a service and i don’t see the timeouts (and 100% cpu core usage).

ibauersachs:
> Any ETA on when this can be merged/released?

Done. It might take a while until it has propagated through Central.

rouzwawi:
Great thanks!

PR #98, opened 2020-03-26
"Record" class access fixes

demonti:
"package" access in the Record class disallows subclasses to reside in other packages. The submitted fix relaxes the access to four methods and a constructor to "protected" access. All required existing subclasses needed to be adjusted as well.


demonti:
Hi Ingo; it is a matter of taste. My personal preference is to apply "final" on classes, methods and member variables only if it is really necessary for some reason (e.g. conceptual/design).

PR #108, opened 2020-05-06
Fix order of OPT and TSIG records in messages

ibauersachs:


demonti:
Hi Ingo,
the changes look good to me. I was a bit puzzled by the AXFR support of SimpleResolver until I realized that it actually issues a zone transfer and then synthesizes a single message response from it.
Klaus

PR #116, opened 2020-07-28
Add support for SVCB and HTTPS records

adam-stoler:
Added functionality and unit tests for handling SVCB and HTTPS records, as described in the following RFC draft -

https://tools.ietf.org/html/draft-ietf-dnsop-svcb-https-01

Please let me know if there are any changes that you would like me to make.  Also, let me know if I should be doing something differently for submitting the Pull Request.  This is my first time contributing to this project.  Thank you.

### DIFF
--- a/src/main/java/org/xbill/DNS/SVCBBase.java
+++ b/src/main/java/org/xbill/DNS/SVCBBase.java
@@ -2,24 +2,26 @@
 package org.xbill.DNS;
 
 import java.io.IOException;
+import java.net.Inet4Address;
+import java.net.Inet6Address;
 import java.net.InetAddress;
 import java.net.UnknownHostException;
 import java.util.ArrayList;
 import java.util.Base64;
 import java.util.HashMap;
+import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.TreeMap;
 import java.util.function.Supplier;
+import java.util.stream.Collectors;
 
-/**
- * Implements common functionality for SVCB and HTTPS records
- */
+/** Implements common functionality for SVCB and HTTPS records */
 abstract class SVCBBase extends Record {
   protected int svcPriority;
   protected Name targetName;
-  protected Map<Integer, ParameterBase> svcParams;
+  protected final Map<Integer, ParameterBase> svcParams;
 
   public static final int MANDATORY = 0;
   public static final int ALPN = 1;
@@ -29,13 +31,23 @@ abstract class SVCBBase extends Record {
   public static final int ECHCONFIG = 5;
   public static final int IPV6HINT = 6;
 
-  protected SVCBBase() {}
+  protected SVCBBase() {
+    svcParams = new TreeMap<>();
+  }
 
   protected SVCBBase(Name name, int type, int dclass, long ttl) {
     super(name, type, dclass, ttl);
+    svcParams = new TreeMap<>();
   }
 
-  protected SVCBBase(Name name, int type, int dclass, long ttl, int priority, Name domain, List<ParameterBase> params) {
+  protected SVCBBase(
+      Name name,
+      int type,
+      int dclass,
+      long ttl,
+      int priority,
+      Name domain,
+      List<ParameterBase> params) {
     super(name, type, dclass, ttl);
     svcPriority = priority;
     targetName = domain;
@@ -97,12 +109,17 @@ abstract class SVCBBase extends Record {
     parameters.add(IPV6HINT, "ipv6hint", ParameterIpv6Hint::new);
   }
 
-  public static abstract class ParameterBase {
+  public abstract static class ParameterBase {
     public ParameterBase() {}
+
     public abstract int getKey();
+
     public abstract void fromWire(byte[] bytes) throws IOException;
+
     public abstract void fromString(String string) throws IOException;
+
     public abstract byte[] toWire();
+
     public abstract String toString();
 
     // Split string on commas, but not if comma is escaped with a '\'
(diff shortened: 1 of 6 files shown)


ibauersachs:
Thanks! I need the next weekend to look at it in detail. In the meantime, could you please apply the license header as follows?
```
// SPDX-License-Identifier: BSD-2-Clause
```

Also, do you expect changes to the records? It's the first draft for the RFC after all...

enygren:
This is actually the ~9th version of the draft (which has been renamed a bunch) and the wire-format should be stablized.  We have an early code-point allocation and WGLC should be happening soon.
Version history: https://datatracker.ietf.org/doc/draft-ietf-dnsop-svcb-https/

Other implementations for interop testing:  https://github.com/MikeBishop/dns-alt-svc/blob/master/svcb-implementations.md

adam-stoler:
I made the update to add the license headers.  I asked Erik (one of the authors of the RFC) about the stability of the current draft and he commented for himself above.  Sounds like it should be fairly stable at this point.

I will be on vacation next week, so I will not be able to respond to anything or make updates from your review until the following week.  Thanks!

ibauersachs:
I only looked at the version number of the current draft, not the entire history. Thanks for the pointer!

There are a few things that must be addressed before I can merge this:
- The build now fails because of invalid formatting (the project uses google-java-format). Please run the formatter using `mvn fmt:format`. To avoid the wrapping in the `assertThrows` in the test files, remove the unnecessary braces from the lambdas.
- Please use `final` in SVCBBase for private fields where possible
- The `Parameter*`-classes seem to provide values that clients will be using, e.g. IP-addresses. Please provide them as their actual type, not as byte[]-blobs (e.g. `ParamterIpv4Hint`: `List<Inet4Address> getAddresses()`)

adam-stoler:
Sorry for the delay.  I was on vacation last week.

I believe the latest commit to my branch has addressed the prior comments.  I tried to use final for private fields where possible.  I was following the pattern of other Record types with functions that parse data into fields of an existing object (as opposed to creating a new one), so I could not do this too widely.

src/main/java/org/xbill/DNS/SVCBBase.java
```diff
@@ -185,20 +204,20 @@ public String toString() {
...
     public ParameterAlpn() {
       super();
       values = new ArrayList<>();
     }
 
-    public ParameterAlpn(List<byte[]> values) {
+    public ParameterAlpn(List<String> values) {
       super();
-      this.values = values;
+      this.values = values.stream().map(String::getBytes).collect(Collectors.toList());
```
ibauersachs, line 216: getBytes uses the platform default encoding. This is often UTF-8, but it's definitely not guaranteed, so unless it's always ASCII, there will be problems.

src/main/java/org/xbill/DNS/SVCBBase.java
```diff
@@ -185,20 +204,20 @@ public String toString() {
...
     }
 
-    public List<byte[]> getValues() {
-      return values;
+    public List<String> getValues() {
+      return values.stream().map(String::new).collect(Collectors.toList());
```
ibauersachs, line 220: new uses the platform default encoding. This is often UTF-8, but it's definitely not guaranteed, so unless it's always ASCII, there will be problems.
(Also for getBytes, Unicode can come from the application and would then result in a mess in binary)

ibauersachs:
> Sorry for the delay. I was on vacation last week.

No worries, I hope you could enjoy them :-)

> I believe the latest commit to my branch has addressed the prior comments. I tried to use final for private fields where possible. I was following the pattern of other Record types with functions that parse data into fields of an existing object (as opposed to creating a new one), so I could not do this too widely.

That's fine with the final, I didn't mean that you need to change code flow, only to make fields final that weren't mutated anyway.

src/main/java/org/xbill/DNS/SVCBBase.java
adam-stoler: Good point.  I made another commit to change this to s.getBytes(StandardCharsets.UTF_8) and new String(b, StandardCharsets.UTF_8).  Is that what you were suggesting?

ibauersachs: Well, yes, if the RFC/draft actually uses UTF-8 for the encoding of ALPNs?

Not so important, but I'd probably store the values as List<String> rather than List<byte[]> and convert in fromWire()/toWire()

adam-stoler: Upon further reading, I think it is limited to just ASCII.  So I will switch to that.

As for storing as a List<byte[]>, I was following the pattern used for other record types, such a TXTBase and AAAARecord, that internally store data this way, presumably to be closer to the wire format.

ibauersachs:
Merged, thanks!

adam-stoler:
Great, thank you!

One more thing, were you planning to or would you be able to do a release soon with these changes and publish it to the Maven Central Repository?  We (at Akamai) would like to start supporting these new record types in our apps.  It is not urgent since we can temporarily use the changes from my fork if necessary.  But I think it is simpler and safer to use the main upstream version from the central repository if possible.  Thanks!

ibauersachs:
I would usually wait a while to see if other changes come along as I'll try to strictly adhere to Semver and don't want to release majro/minor versions too frequently.
I could release the snapshots to Maven Central, but I guess that wouldn't help you much?

adam-stoler:
Sorry for the delay, I had to check with the rest of the team regarding this.  Releasing the snaphots to Maven Central would actually be useful to us for integration.  So if it is not too much trouble, we would appreciate it if you could do that.  Otherwise, we are fine with picking this up on the normal release schedule.  Thanks!

PR #119, opened 2020-08-21
Svcb https.fixes

adam-stoler:
Further code review and interoperability testing identified a potential issue with the ALPN parameter values for SVCB and HTTPS records.  The RFC appears to allow (or at least does not preclude) the use of escaped byte sequences in the ALPN values, and other implementations of SVCB/HTTPS support it.  This change adds that support to dnsjava and additional unit tests to exercise it.

### DIFF
--- a/src/test/java/org/xbill/DNS/SVCBRecordTest.java
+++ b/src/test/java/org/xbill/DNS/SVCBRecordTest.java
@@ -163,7 +163,8 @@ public class SVCBRecordTest {
     String str = "1 . alpn=\"h2\\,h3,h\\\\4\"";
     String expectedStr = "1 . alpn=h2\\,h3,h\\\\4";
     byte[] bytes = stringToWire(str);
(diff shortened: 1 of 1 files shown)


adam-stoler:
Sounds great, thank you!

Should I be able to see a 3.3.0-SNAPSHOT version in Maven Central now?  I still see 3.2.2 as the latest version here -https://search.maven.org/artifact/dnsjava/dnsjava.  I don't see an indication of the auto publish in the build logs - https://travis-ci.org/github/dnsjava/dnsjava/builds/720239474

ibauersachs:
The publishing is part of the Github Actions, not Travis. And the Maven Central search doesn't show snapshots. If you just specify 3.3.0-SNAPSHOT as a dependency you should be able to use it.

adam-stoler:
Ah, I see.  Thanks for clarifying.  I think the issue might be that our internal nexus does not proxy the snapshot repository.  We will do our own integration until a new release is available.

PR #120, opened 2020-08-24
Support IPv4 mapped IPv6 address in AAAA record

spwei:
1. let AAAARecord support IPv4 Address.
2. use ::ffff:x.x.x.x instead of 0:0:0:0:ffff:xx:xx, more readable

### DIFF
--- a/src/test/java/org/xbill/DNS/AAAARecordTest.java
+++ b/src/test/java/org/xbill/DNS/AAAARecordTest.java
@@ -87,13 +87,9 @@ class AAAARecordTest {
   }
 
   @Test
-  void ctor_v4() {
-    try {
+  void ctor_v4() throws UnknownHostException {
       AAAARecord ar = new AAAARecord(m_an, DClass.IN, m_ttl, InetAddress.getByName("192.168.1.1"));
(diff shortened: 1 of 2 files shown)


src/test/java/org/xbill/DNS/AAAARecordTest.java
```diff
@@ -85,14 +84,15 @@ void ctor_4arg() {
...
+  void ctor_v4() {
     try {
-      new AAAARecord(m_an, DClass.IN, m_ttl, InetAddress.getByName("192.168.0.1"));
-      fail("IllegalArgumentException not thrown");
-    } catch (IllegalArgumentException e) {
-    } catch (UnknownHostException e) {
-      fail(e.getMessage());
+      AAAARecord ar = new AAAARecord(m_an, DClass.IN, m_ttl, InetAddress.getByName("192.168.1.1"));
+      assertEquals("::ffff:192.168.1.1", ar.rrToString());
+    } catch (UnknownHostException ignore) {
```
ibauersachs, line 94: Please add a `throws UnknownHostException` at the method level instead silently catching it (event if it can't happen).

ibauersachs (changes requested):

Issue #123, opened 2020-09-02
DOH resolver to support persistent connection


devrivne:
From what I'm seeing, DOH resolver creates a new connection on each lookup that issues to:
With light load
```
$ netstat -natp | grep here-doh-server-ip  |wc -l
865
```

With a slihtly increased load:
```
$ sudo netstat -natp | grep here-doh-server-ip  |wc -l
7120
```

Logs:
```
java.util.concurrent.CompletionException: java.net.http.HttpConnectTimeoutException: HTTP connect timed out
java.util.concurrent.CompletionException: java.io.IOException: too many concurrent streams
[86.730s][warning][os,thread] Failed to start thread - pthread_create failed (EAGAIN) for attributes: stacksize: 1024k, guardsize: 0k, detached.
```

Also, is it possible to add connection pool?

ibauersachs:
Which version of Java did you use for this test? If it is Java 8, then I won't bother about it as this cannot use the recommended HTTP v2.

devrivne:
It was tested on on 11 and 8 versions.

ibauersachs:
I did a quick test and cannot reproduce this. HTTP keepalive is enabled by default on Java 11 (and 8 as well). You need to reuse the `DohResolver` instance though. If you create `DohResolver` per request, it will create a new [HttpClient](https://docs.oracle.com/en/java/javase/11/docs/api/java.net.http/java/net/http/HttpClient.html) and thus new threads and new connections. `DohResolver` is thread safe unless you use TSIG.

devrivne:
There is no sense in creating a resolver per request.
```
val resolver = new org.xbill.DNS.DohResolver("https://cloudflare-dns.com/dns-query")
resolver.setTimeout(timeout)
resolver.sendAsync(message).toCompletableFuture()
```

Trying to resolve from 200 up to 1000 domains per second asynchronously.

ibauersachs:
**Java 11**
[...]

PR #127, opened 2020-09-08
Allow for an unknown content length.

twalmsley:
My DoH service works fine when accessed directly, but when it is behind an API gateway the response content length is unknown, so `conn.getContentLength()` returns `-1` which causes an exception to be thrown.

I have 2 suggested changes here:

1. When the content length is unknown, just create a fixed size buffer for the response. I have chosen a maximum size that seems sensible to me, and I know I could have made it dynamic, but I think even in that case there would still be a need for a sensible maximum buffer size. I can change this if needed.
1. When reading from the connection stream, it is always trying to read `responseBytes.length` bytes, but if it has already read some bytes then there will be no room for that number of bytes, so I have used the `offset` to reduce this as bytes are read into the buffer. I realise that if the content length is known then it won't read more than the expected number of bytes, but in the case when the length is not known it will be possible to read more bytes than the allocated buffer size, hence the check.

I have tested this against `https://dns.google/dns-query` which does return a valid content length, as well as my own service behind an API gateway which returns -1 for the content length, and it succeeds in both cases.

### DIFF
--- a/src/main/java/org/xbill/DNS/DohResolver.java
+++ b/src/main/java/org/xbill/DNS/DohResolver.java
@@ -9,6 +9,7 @@ import java.net.HttpURLConnection;
 import java.net.URI;
 import java.net.URL;
 import java.time.Duration;
+import java.util.Arrays;
 import java.util.List;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionStage;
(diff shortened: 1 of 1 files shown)


ibauersachs:
Thanks, I've worked on the DohResolver for #123 and implemented this slightly different as you cannot assume that 2048bytes are enough, messages with dnssec easily exceeds this. I couldn't really test it, so please let me know if master works for you as well.

twalmsley:
Great, thanks for making the change - I'll look out for release 3.3 and give it a try, but it looks like a good fix to me.

PR #153, opened 2021-02-02
Use fail() from junit

nresare:
The previous import was from the junit 4 classes that ships with robolectric, which seems like it was a mistake.


ibauersachs (approved):
Good catch!

PR #154, opened 2021-02-02
Add available error information to UnknownHostException

### LINKED ISSUE #149: Suggestion: add more information to unknown host exceptions
I see that there is additional information in the `Lookup` class about why a lookup failed. Adding that information to the message of the unknown host exception may be useful for those trying to debug DNS issues.

A suggestion is here:
```
final String message = String.format("Unknown host: %s result: %d error string: %s", name,
                        lookup.getResult(), lookup.getErrorString());
throw new UnknownHostException(message);
```

ibauersachs:
Closes #149

### DIFF
--- a/src/main/java/org/xbill/DNS/Address.java
+++ b/src/main/java/org/xbill/DNS/Address.java
@@ -267,7 +267,8 @@ public final class Address {
             return aaaa;
           }
         }
-        throw new UnknownHostException("<" + name + "> could not be resolved");
+        throw new UnknownHostException(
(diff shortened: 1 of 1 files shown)


nresare (approved):
I have a weak preference for using colon instead of semicolon to separate failure from reason as it s in line with for example vintage unix tooling:
```bash
noa@Noas-MacBook-Air dnsjava % ls /foo 
ls: /foo: No such file or directory
```

ibauersachs:
Updated to colons, I have no preference.

nresare:
👍

PR #155, opened 2021-02-02
Run only on master and PRs, and Sonar only on Ubuntu/x64/Java11

ibauersachs:
Remove Travis and Coveralls:
- Travis regularly times out and is already replaced with GitHub Actions anyway
- Coveralls doesn't provide proper GitHub Actions support (i.e. coverage reports of PRs from forks), that's why Travis was still there

Add Sonar:
- Get a some better static analysis reports and coverage
- Add codecov as well because Sonar also doesn't support analysing PRs from forks so that there's at least a coverage report

Add checkstyle:
To minimally check g-j-f stuff that isn't in the formatter jar (the whole google_checks.xml cannot be applied because of the legacy package name and various class names which we don't want to rename for backwards compatibility).


nresare (approved):
Although I am not familiar with the specifics of the services and tools you are using, I am all for static analysis and code formatting tools. Reading through the changes I it looks vaguely sensible. 👍 I shall take this as an opportunity to get into the habit to push my PR branches to the main repo.

PR #156, opened 2021-02-03
WIP: move the Lookup state machine into a separate class

nresare:
In preparation for some refactoring of the Lookup.run() logic
this change moves run(), and the various result getters
to a separate non-public class LookupHelper. This change
is not intended to introduce any behaviour changes.

### DIFF
--- a/src/main/java/org/xbill/DNS/Lookup.java
+++ b/src/main/java/org/xbill/DNS/Lookup.java
@@ -2,8 +2,6 @@
 
 package org.xbill.DNS;
 
-import java.io.IOException;
-import java.io.InterruptedIOException;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashMap;
@@ -34,6 +32,8 @@ public final class Lookup {
   private static List<Name> defaultSearchPath;
   private static Map<Integer, Cache> defaultCaches;
   private static int defaultNdots;
+  private final LookupHelper lookupHelper;
+  private final int dclass;
 
   private Resolver resolver;
   private List<Name> searchPath;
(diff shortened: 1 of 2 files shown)


nresare:
Mostly opening this PR to have a starting point for a discussion. I understand that the change is somewhat opinionated, please do tell if it goes against your sensibilities.

src/main/java/org/xbill/DNS/Lookup.java
```diff
@@ -449,212 +395,30 @@ public void setCycleResults(boolean cycleResults) {
...
-    if (done && result != -1) {
-      return;
+    if (resolver == null) {
+      resolver = getDefaultResolver();
     }
-    StringBuilder sb = new StringBuilder("Lookup of " + name + " ");
-    if (dclass != DClass.IN) {
-      sb.append(DClass.string(dclass)).append(" ");
+    if (searchPath == null) {
+      getDefaultSearchPath();
```
ibauersachs, line 411: Missing assignment?

ibauersachs (changes requested):
I don't quite yet see the point of it. What are you trying to achieve? With this change, but also with what you're planning in run().

nresare:
Thank you for having a look, and for finding the missing assignment :)

I agree that this change doesn't bring much value in and of itself, but it kind of fits into a bigger picture in my head.  My thinking goes a bit like this:

* The actual lookup implementation in Lookup has some unconventional use of shared state instead of passing return values around. It requires the reader of the code to keep a lot of state in their head to be able to follow what is going on.
* I first step towards disentangling some of that would be to increase the unit test coverage, to make sure that refactoring can be done without changing the behaviour of the public interfaces.
* When I started looking at that, I realised that Lookup essentially does two things. One is a (very unconventional) factory type construct that persists defaults deals with configuration and the other part is the lookup implementation. 
* When starting to change the code, I realise that I would like to propose a new interface to making queries where the caller of the code would hold some sort of long lived instance that managed configuration and then issue queries by calling a method on that long lived instance.  At the same time I expect the current way of doing things to be around for a long time. To be able to experiment with this change it would be useful to have the lookup algorithm be split out.

In hindsight, it might make sense to simply let this change be for a bit, and wait until I have a proposal for the new interface before I introduce this split.

ibauersachs:
Okay, just go ahead, I don't mind the changes. The class just wasn't bad enough that I felt it was necessary to mess around too much yet.
If you design a new interface, you might want to make that thing properly async with `CompletionStage<>`s. If you don't want this merged beforehand, maybe convert the PR to a draft.

PR #158, opened 2021-02-03
Increase test coverage in LookupTest

nresare:
Trying to hit as many code paths in the Lookup.run() as possible. Seems prudent before refactoring :)

### DIFF
--- a/src/test/java/org/xbill/DNS/LookupTest.java
+++ b/src/test/java/org/xbill/DNS/LookupTest.java
@@ -2,6 +2,8 @@
 package org.xbill.DNS;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNull;
+import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.Mockito.times;
 import static org.mockito.Mockito.verify;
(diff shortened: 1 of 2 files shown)


src/test/java/org/xbill/DNS/LookupTest.java
```diff
@@ -2,89 +2,345 @@
...
+    wireUpMockResolver(mockResolver, this::simpleAnswer);
+
+    Lookup lookup = makeLookupWithResolver(mockResolver, "host");
+    lookup.setSearchPath("example.com");
+    Record[] results = lookup.run();
+
+    ArgumentCaptor<Message> messageCaptor = ArgumentCaptor.forClass(Message.class);
+    verify(mockResolver).send(messageCaptor.capture());
+
+    assertEquals(
```
ibauersachs, line 81: The default ndots is 1, a lookup for "host" should thus initially be made relative, i.e. without search path. I'm quite surprised this test passed sometimes, but it correctly failed on the Windows test run.

src/test/java/org/xbill/DNS/LookupTest.java
```diff
@@ -2,89 +2,345 @@
...
+          return new CNAMERecord(
+              DUMMY_NAME, DClass.IN, 0, Name.fromConstantString("first.example.com."));
+        });
+  }
+
+  @Test
+  void testRun_reuseLookup() throws Exception {
+    wireUpMockResolver(mockResolver, this::simpleAnswer);
+
+    Lookup lookup = makeLookupWithResolver(mockResolver, "host");
```
ibauersachs, line 171: Same here, ndots at 1 means relative lookup. There might be more, I didn't thoroughly look at all tests.

ibauersachs (changes requested):

src/test/java/org/xbill/DNS/LookupTest.java
```diff
@@ -2,89 +2,345 @@
 package org.xbill.DNS;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNull;
 import static org.mockito.ArgumentMatchers.any;
-import static org.mockito.Mockito.times;
-import static org.mockito.Mockito.verify;
-import static org.mockito.Mockito.when;
+import static org.mockito.Mockito.*;
```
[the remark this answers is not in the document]
nresare: I am no fan of wildcard imports. This is what default settings intellij with `mvn com.coveo:fmt-maven-plugin:2.9.1:format` applied afterwards came up with.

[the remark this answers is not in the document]
ibauersachs: Well, seems the formatter does not want to handle that (google/google-java-format#113). Can you please manually configure IntelliJ to not use wildcards as documented here:
https://github.com/google/google-java-format#intellij-android-studio-and-other-jetbrains-ides

I hate to bring checkstyle back for that :-(

src/test/java/org/xbill/DNS/LookupTest.java
nresare: That is not how I read Lookup.java:601-603. A Name with the naked value `host` has `labels() == 1` and as such it doesn't match the conditional looking for labels() values more than 1 in the default case.

My understanding of the default value of `ndots` is that most people want A to be interpreted relative to the search path but they want A.B to be interpreted as a fully qualified Name. However, some people might have a more complicated setup, with hosts in `DC_A` and `DC_B`: thus they want to be able to set `ndots` to a higher value and be able to resolve `HOST.DC_A` as something relative to search path.

With this in mind it seems reasonable to expect a lookup for a naked `host` with `ndots == 1` to first be queried with each of the search path appended to it.

src/test/java/org/xbill/DNS/LookupTest.java
```diff
@@ -2,89 +2,345 @@
...
+        messageCaptor.getValue().getSection(Section.QUESTION).get(0));
+
+    assertEquals(1, results.length);
+  }
+
+  @Test
+  void testNdots1() throws Exception {
+    Resolver mockResolver = Mockito.mock(Resolver.class);
+    wireUpMockResolver(mockResolver, this::simpleAnswer);
+    Lookup l = makeLookupWithResolver(mockResolver, "example.com");
```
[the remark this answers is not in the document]
nresare: Good catch, it was there in an earlier version. I'll re-add it.

src/test/java/org/xbill/DNS/LookupTest.java
ibauersachs: Yes, you're right. I always get confused by the wording of resolv.conf's manual:

> ndots:n
sets a threshold for the number of dots which must appear in a name given to res_query(3) (see resolver(3)) before an initial absolute query will be made. The default for n is 1, meaning that if there are any dots in a name, the name will be tried first as an absolute name before any search list elements are appended to it. The value for this option is silently capped to 15.

Anyway, something's wrong somewhere (not necessarily your new tests), as the tests fail when run on Windows.

nresare: I'll dig out an intel laptop and a  Window VM and see if I can reproduce the problem and get to the bottom of this

ibauersachs: Got an M1 as a parting gift?
If you push directly to dnsjava I can check out your branches more easily.

nresare: It turns out that the `WindowsResolverConfigProvider` sets ndots to -1 to indicate that it does not support that configuration, so this failure uncovered a real customer facing issue. Yay!

src/main/java/org/xbill/DNS/Lookup.java
```diff
@@ -268,7 +268,8 @@ public Lookup(Name name, int type, int dclass) {
       this.searchPath = getDefaultSearchPath();
       this.cache = getDefaultCache(dclass);
     }
-    this.ndots = defaultNdots;
+    // negative defaultsNdots means that the config provider did not support the config type
+    this.ndots = defaultNdots == -1 ? 1 : defaultNdots;
```
ibauersachs, line 272: Very good finding! I wonder though, maybe it would be better to default to 1 instead of -1 in the `ResolverConfig` interface, especially if we would default to 1 here anyway.

And something else: the tests set `CONFIGPROVIDER_SKIP_INIT`, so why was the `WindowsResolverConfig` used anyway?

nresare: I would be happy to simplify `ResolverConfigProvider` to return the default if the value is not set.

I believe that `CONFIGPROVIDER_SKIP_INIT` fails to prevent WindowsResolverConfigProvider due to the fact that some other test triggers it before `LookupTest` is run. I have verified that the before() and after() methods gets properly run with a debugger

ibauersachs (approved):

PR #161, opened 2021-02-09
began experimenting with DNSSession

nresare:
I have started experimenting with an asynchronous lookup mechanism. Please note that this is a draft and not not feature complete. I just want to give the opportunity to gather some early feedback. In particular I am curious to to hear about:

* The name `DNSSession` for the new long lived API endpoint. Session is kind of vague on purpose, to allow it to have other public methods in the future that interact with DNS servers. I want to avoid `Session` because of the inconvenience of having it clash with for example the Session classes in the Neo4j or Cassandra.
* What do we think about putting an exception hierarchy in a separate module? My main reason for going there is that the org.xbill.DNS module is pretty crowded, but this is a weak preference.

The idea is to implement the following:
* Add aliases to QueryResult
* caching
* search path expansion
*  ndots handling
* the cycleResults feature
* a configurable minimum credibility requirement

Once this is in place, this query implementation would reach feature parity with Lookup.run() and we could replace that one with this without changing the public API (besides making it Deprecated, possibly).

### DIFF
--- a/src/main/java/org/xbill/DNS/DNSSession.java
+++ b/src/main/java/org/xbill/DNS/DNSSession.java
@@ -1,0 +1,137 @@
+package org.xbill.DNS;
+
+import static java.lang.String.format;
+
+import java.util.List;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionStage;
+import org.xbill.DNS.exceptions.AdditionalDetail;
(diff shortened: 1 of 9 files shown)


nresare:
Closing this and re-opening against a branch on the dnsjava repo to see if the static code analysis is more friendly to that.

PR #162, opened 2021-02-09
Introduce an asynchronous lookup API

nresare:
I have started experimenting with an asynchronous lookup mechanism. Please note that this is a draft and not not feature complete. I just want to give the opportunity to gather some early feedback. In particular I am curious to to hear about:

* The name `DNSSession` for the new long lived API endpoint. Session is kind of vague on purpose, to allow it to have other public methods in the future that interact with DNS servers. I want to avoid Session because of the inconvenience of having it clash with for example the Session classes in the Neo4j or Cassandra.
* What do we think about putting an exception hierarchy in a separate module? My main reason for going there is that the org.xbill.DNS module is pretty crowded, but this is a weak preference.

The idea is to implement the following:
* Add aliases to QueryResult
* caching
* search path expansion
* ndots handling
* the cycleResults feature
* a configurable minimum credibility requirement
Once this is in place, this query implementation would reach feature parity with Lookup.run() and we could replace that one with this without changing the public API (besides making it Deprecated, possibly).

### DIFF
--- a/src/main/java/org/xbill/DNS/lookup/InvalidZoneDataException.java
+++ b/src/main/java/org/xbill/DNS/lookup/InvalidZoneDataException.java
@@ -1,0 +1,13 @@
+// SPDX-License-Identifier: BSD-2-Clause
+package org.xbill.DNS.lookup;
+
+/**
+ * Sometimes DNS zone data involved in the lookup might be violating specifications. For example, a
+ * DNAME expansion might result in names that are too long or a query response might hold multiple
+ * CNAME records.
+ */
+public class InvalidZoneDataException extends LookupFailedException {
+  public InvalidZoneDataException(String message) {
+    super(message);
+  }
+}
--- a/src/main/java/org/xbill/DNS/lookup/LookupResult.java
+++ b/src/main/java/org/xbill/DNS/lookup/LookupResult.java
@@ -1,0 +1,40 @@
+// SPDX-License-Identifier: BSD-2-Clause
+package org.xbill.DNS.lookup;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import lombok.EqualsAndHashCode;
+import lombok.Getter;
+import lombok.ToString;
+import org.xbill.DNS.Name;
+import org.xbill.DNS.Record;
+
+/** LookupResult instances holds the result of a successful lookup operation. */
+@ToString
+@EqualsAndHashCode
+public final class LookupResult {
+  /** An unmodifiable list of records that this instance wraps, may not be null but can be empty */
+  @Getter private final List<Record> records;
+  /**
+   * In the case of CNAME or DNAME indirection, this property contains the original name as well as
+   * any intermediate redirect targets except the last one. For example, if X is a CNAME pointing to
+   * Y which is a CNAME pointing to Z which has an A record, aliases will hold X and Y after
+   * successful lookup.
+   */
+  @Getter private List<Name> aliases;
+
+  /**
+   * Construct an instance with the provided records and, in the case of a CNAME or DNAME
+   * indirection a List of aliases.
+   *
+   * @param records a list of records to return.
+   * @param aliases a list of aliases discovered during lookup, or null if there was no indirection.
+   */
+  public LookupResult(List<Record> records, List<Name> aliases) {
+    this.records = Collections.unmodifiableList(new ArrayList<>(records));
+    if (aliases != null) {
+      this.aliases = Collections.unmodifiableList(new ArrayList<>(aliases));
+    }
+  }
+}
--- a/src/main/java/org/xbill/DNS/lookup/LookupSession.java
+++ b/src/main/java/org/xbill/DNS/lookup/LookupSession.java
@@ -1,0 +1,310 @@
+// SPDX-License-Identifier: BSD-2-Clause
+package org.xbill.DNS.lookup;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionStage;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+import lombok.Builder;
+import lombok.NonNull;
+import lombok.Singular;
+import org.xbill.DNS.CNAMERecord;
+import org.xbill.DNS.Cache;
+import org.xbill.DNS.Credibility;
+import org.xbill.DNS.DClass;
+import org.xbill.DNS.DNAMERecord;
+import org.xbill.DNS.Message;
+import org.xbill.DNS.Name;
+import org.xbill.DNS.NameTooLongException;
+import org.xbill.DNS.Rcode;
+import org.xbill.DNS.Record;
+import org.xbill.DNS.Resolver;
+import org.xbill.DNS.Section;
+import org.xbill.DNS.SetResponse;
+import org.xbill.DNS.Type;
+
+/**
+ * LookupSession provides facilities to make DNS Queries. A LookupSession is intended to be long
+ * lived, and it's behaviour can be configured using the properties of the LookupSessionBuilder
+ * instance returned by the builder() method.
+ */
+@Builder
+public class LookupSession {
+  public static final int DEFAULT_MAX_ITERATIONS = 16;
+  public static final int DEFAULT_NDOTS = 1;
+  /** The {@link Resolver} to use to look up records. */
+  @NonNull private final Resolver resolver;
+  /**
+   * Sets the maximum number of CNAME or DNAME redirects allowed before lookups with fail with
+   * {@link RedirectOverflowException}. Defaults to {@value
+   * org.xbill.DNS.lookup.LookupSession#DEFAULT_MAX_ITERATIONS}.
+   */
+  @Builder.Default private final int maxRedirects = DEFAULT_MAX_ITERATIONS;
+  /**
+   * Sets the threshold for the number of dots which must appear in a name before it is considered
+   * absolute. The default is {@value org.xbill.DNS.lookup.LookupSession#DEFAULT_NDOTS}, meaning
+   * that if there are any dots in a name, the name will be tried first as an absolute name.
+   */
+  @Builder.Default private final int ndots = DEFAULT_NDOTS;
+
+  /** Configures the search path used to look up relative names with less than ndots dots. */
+  @Singular("searchPath")
+  private final List<Name> searchPath;
+
+  /**
+   * If set to {@code true}, cached results with multiple records will be returned with the starting
+   * point shifted one step per request.
+   */
+  @Builder.Default private final boolean cycleResults = false;
+
+  /**
+   * Configures the Cache instances to be used for lookups for the different {@link DClass} values.
+   */
+  @Singular("cache")
+  private final Map<Integer, Cache> caches;
+
+  /**
+   * A builder for {@link LookupSession} instances where functionality is mostly generated as
+   * described in the <a href="https://projectlombok.org/features/Builder">Lombok Builder</a>
+   * documentation. An instance of this class is obtained by calling {@link LookupSession#builder()}
+   * and configured using the methods with names corresponding to the different properties. Once
+   * fully configured, a {@link LookupSession} instance is obtained by calling {@link
+   * LookupSessionBuilder#build()} on the builder instance.
+   */
+  public static class LookupSessionBuilder {
+    void preBuild() {
+      // note that this transform is idempotent, as concatenating an already absolute Name with root
+      // is a noop.
+      if (searchPath != null) {
+        this.searchPath =
+            searchPath.stream()
+                .map(
+                    name -> {
+                      try {
+                        return Name.concatenate(name, Name.root);
+                      } catch (NameTooLongException e) {
+                        throw new IllegalArgumentException("Search path name too long");
+                      }
+                    })
+                .collect(Collectors.toCollection(ArrayList::new));
+      }
+    }
+  }
+
+  /** Returns a new {@link LookupSessionBuilder} instance. */
+  public static LookupSessionBuilder builder() {
+    return new LookupSessionBuilder() {
+      @Override
+      public LookupSession build() {
+        preBuild();
+        return super.build();
+      }
+    };
+  }
+
+  /**
+   * Make an asynchronous lookup of the provided name.
+   *
+   * @param name the name to look up.
+   * @param type the type to look up, values should correspond to constants in {@link Type}.
+   * @param dclass the class to look up, values should correspond to constants in {@link DClass}.
+   * @return A {@link CompletionStage} what will yield the eventual lookup result.
+   */
+  public CompletionStage<LookupResult> lookupAsync(Name name, int type, int dclass) {
+    CompletableFuture<LookupResult> future = new CompletableFuture<>();
+    lookupUntilSuccess(expandName(name).iterator(), type, dclass, future);
+    return future;
+  }
+
+  /**
+   * Generate a stream of names according to the search path application semantics. The semantics of
+   * this is a bit odd, but they are inherited from Lookup.java. Note that the stream returned is
+   * never empty, as it will at the very least always contain {@code name}.
+   */
+  Stream<Name> expandName(Name name) {
+    if (name.isAbsolute()) {
+      return Stream.of(name);
+    }
+    Stream<Name> fromSearchPath =
+        Stream.concat(
+            searchPath.stream()
+                .map(searchSuffix -> safeConcat(name, searchSuffix))
+                .filter(Objects::nonNull),
+            Stream.of(safeConcat(name, Name.root)));
+
+    if (name.labels() > ndots) {
+      return Stream.concat(Stream.of(safeConcat(name, Name.root)), fromSearchPath);
+    }
+    return fromSearchPath;
+  }
+
+  private static Name safeConcat(Name name, Name suffix) {
+    try {
+      return Name.concatenate(name, suffix);
+    } catch (NameTooLongException e) {
+      return null;
+    }
+  }
+
+  private void lookupUntilSuccess(
+      Iterator<Name> names, int type, int dclass, CompletableFuture<LookupResult> future) {
+
+    Record query = Record.newRecord(names.next(), type, dclass);
+    lookupWithCache(query, null)
+        .thenCompose(answer -> resolveRedirects(answer, query))
+        .whenComplete(
+            (result, ex) -> {
+              Throwable cause = ex == null ? null : ex.getCause();
+              if (cause instanceof NoSuchDomainException || cause instanceof NoSuchRRSetException) {
+                if (names.hasNext()) {
+                  lookupUntilSuccess(names, type, dclass, future);
+                } else {
+                  future.completeExceptionally(cause);
+                }
+              } else if (cause != null) {
+                future.completeExceptionally(cause);
+              } else {
+                future.complete(result);
+              }
+            });
+  }
+
+  private CompletionStage<LookupResult> lookupWithCache(Record queryRecord, List<Name> aliases) {
+    return Optional.ofNullable(caches.get(queryRecord.getDClass()))
+        .map(c -> c.lookupRecords(queryRecord.getName(), queryRecord.getType(), Credibility.NORMAL))
+        .map(setResponse -> setResponseToMessageFuture(setResponse, queryRecord))
+        .orElseGet(() -> lookupWithResolver(queryRecord, aliases));
+  }
+
+  private CompletionStage<LookupResult> lookupWithResolver(Record queryRecord, List<Name> aliases) {
+    return resolver
+        .sendAsync(Message.newQuery(queryRecord))
+        .thenApply(this::maybeAddToCache)
+        .thenApply(answer -> buildResult(answer, aliases, queryRecord));
+  }
+
+  private Message maybeAddToCache(Message message) {
+    Optional.ofNullable(caches.get(message.getQuestion().getDClass()))
+        .ifPresent(cache -> cache.addMessage(message));
+    return message;
+  }
+
+  private CompletionStage<LookupResult> setResponseToMessageFuture(
+      SetResponse setResponse, Record queryRecord) {
+    if (setResponse.isNXDOMAIN()) {
+      return completeExceptionally(
+          new NoSuchDomainException(queryRecord.getName(), queryRecord.getType()));
+    }
+    if (setResponse.isNXRRSET()) {
+      return completeExceptionally(
+          new NoSuchRRSetException(queryRecord.getName(), queryRecord.getType()));
+    }
+    if (setResponse.isSuccessful()) {
+      List<Record> records =
+          setResponse.answers().stream()
+              .flatMap(rrset -> rrset.rrs(cycleResults).stream())
+              .collect(Collectors.toList());
+      return CompletableFuture.completedFuture(new LookupResult(records, null));
+    }
+    return null;
+  }
+
+  private <T extends LookupFailedException> CompletionStage<LookupResult> completeExceptionally(
+      T failure) {
+    CompletableFuture<LookupResult> future = new CompletableFuture<>();
+    future.completeExceptionally(failure);
+    return future;
+  }
+
+  private CompletionStage<LookupResult> resolveRedirects(LookupResult response, Record query) {
+    CompletableFuture<LookupResult> future = new CompletableFuture<>();
+    maybeFollowRedirect(response, query, 1, future);
+    return future;
+  }
+
+  private void maybeFollowRedirect(
+      LookupResult response,
+      Record query,
+      int redirectCount,
+      CompletableFuture<LookupResult> future) {
+    try {
+      if (redirectCount > maxRedirects) {
+        throw new RedirectOverflowException(
+            String.format("Refusing to follow more than %s redirects", maxRedirects));
+      }
+
+      List<Record> records = response.getRecords();
+      if (records.isEmpty()) {
+        future.complete(response);
+      } else if (records.get(0).getType() == Type.DNAME || records.get(0).getType() == Type.CNAME) {
+        lookupWithCache(
+                buildRedirectQuery(response, query),
+                makeAliases(response.getAliases(), query.getName()))
+            .thenAccept(m -> maybeFollowRedirect(m, query, redirectCount + 1, future));
+      } else {
+        future.complete(response);
+      }
+    } catch (LookupFailedException e) {
+      future.completeExceptionally(e);
+    }
+  }
+
+  /** Return an unmodifiable list containing the contents of previous, if any, plus name. */
+  private List<Name> makeAliases(List<Name> previous, Name name) {
+    if (previous == null) {
+      return Collections.singletonList(name);
+    } else {
+      List<Name> copy = new ArrayList<>(previous);
+      copy.add(name);
+      return Collections.unmodifiableList(copy);
+    }
+  }
+
+  private Record buildRedirectQuery(LookupResult response, Record question) {
+    List<Record> answer = response.getRecords();
+    Record firstAnswer = answer.get(0);
+    if (answer.size() != 1) {
+      throw new InvalidZoneDataException("Multiple CNAME RRs not allowed, see RFC1034 3.6.2");
+    }
+
+    if (firstAnswer.getType() == Type.CNAME) {
+      return Record.newRecord(
+          ((CNAMERecord) firstAnswer).getTarget(), question.getType(), question.getDClass());
+    }
+    // if it is not a CNAME, it's a DNAME
+    try {
+      Name name = question.getName().fromDNAME((DNAMERecord) firstAnswer);
+      return Record.newRecord(name, question.getType(), question.getDClass());
+    } catch (NameTooLongException e) {
+      throw new InvalidZoneDataException(
+          "DNAME redirect would result in a name that would be too long");
+    }
+  }
+
+  /** Returns a LookupResult if this response was a non-exceptional empty result, else null. */
+  private static LookupResult buildResult(Message answer, List<Name> aliases, Record query) {
+    int rcode = answer.getRcode();
+    List<Record> answerRecords = answer.getSection(Section.ANSWER);
+    if (answerRecords.isEmpty()) {
+      switch (rcode) {
+        case Rcode.NXDOMAIN:
+          throw new NoSuchDomainException(query.getName(), query.getType());
+        case Rcode.NXRRSET:
+          throw new NoSuchRRSetException(query.getName(), query.getType());
+        case Rcode.SERVFAIL:
+          throw new ServerFailedException();
+        default:
+          throw new LookupFailedException(
+              String.format("Unknown non-success error code %s", Rcode.string(rcode)));
+      }
+    }
+    return new LookupResult(answerRecords, aliases);
+  }
+}
--- a/src/main/java/org/xbill/DNS/lookup/RedirectOverflowException.java
+++ b/src/main/java/org/xbill/DNS/lookup/RedirectOverflowException.java
@@ -1,0 +1,12 @@
+// SPDX-License-Identifier: BSD-2-Clause
+package org.xbill.DNS.lookup;
+
+/**
+ * Thrown if the lookup results in too many CNAME and/or DNAME indirections. This would be the case
+ * for example if two CNAME records point to each other.
+ */
+public class RedirectOverflowException extends LookupFailedException {
+  public RedirectOverflowException(String message) {
+    super(message);
+  }
+}
--- a/lombok.config
+++ b/lombok.config
@@ -1,0 +1,1 @@
+lombok.addLombokGeneratedAnnotation = true
--- a/pom.xml
+++ b/pom.xml
@@ -53,6 +53,7 @@
         <sonar.host.url>https://sonarcloud.io</sonar.host.url>
         <sonar.java.source>8</sonar.java.source>
         <sonar.coverage.jacoco.xmlReportPaths>${project.build.directory}/site/jacoco/jacoco.xml</sonar.coverage.jacoco.xmlReportPaths>
+        <delombok.output>${project.build.directory}/generated-sources/delombok</delombok.output>
     </properties>
 
     <build>
@@ -141,6 +142,7 @@
                     <windowtitle>dnsjava documentation</windowtitle>
                     <footer/>
                     <doclint>none</doclint>
+                    <sourcepath>${delombok.output}</sourcepath>
                 </configuration>
             </plugin>
 
@@ -302,6 +304,25 @@
                     </execution>
                 </executions>
             </plugin>
+
+            <plugin>
+                <groupId>org.projectlombok</groupId>
+                <artifactId>lombok-maven-plugin</artifactId>
+                <version>1.18.18.0</version>
+                <configuration>
+                    <sourceDirectory>${project.build.sourceDirectory}</sourceDirectory>
+                    <addOutputDirectory>false</addOutputDirectory>
+                </configuration>
+                <executions>
+                    <execution>
+                        <phase>generate-sources</phase>
+                        <goals>
+                            <goal>delombok</goal>
+                        </goals>
+                    </execution>
+                </executions>
+            </plugin>
+
         </plugins>
     </build>
 
@@ -360,6 +381,12 @@
             <scope>test</scope>
         </dependency>
         <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-junit-jupiter</artifactId>
+            <version>3.6.0</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
             <groupId>org.slf4j</groupId>
             <artifactId>slf4j-simple</artifactId>
             <version>${slf4j.version}</version>
@@ -395,7 +422,8 @@
                         <groupId>org.apache.maven.plugins</groupId>
                         <artifactId>maven-javadoc-plugin</artifactId>
                         <configuration>
-                            <excludePackageNames>*.spi.*</excludePackageNames>
+                            <excludePackageNames>*.spi</excludePackageNames>
+                            <sourcepath>${delombok.output}</sourcepath>
                         </configuration>
                     </plugin>
                 </plugins>
--- a/src/main/java/org/xbill/DNS/lookup/LookupFailedException.java
+++ b/src/main/java/org/xbill/DNS/lookup/LookupFailedException.java
@@ -1,0 +1,34 @@
+// SPDX-License-Identifier: BSD-2-Clause
+package org.xbill.DNS.lookup;
+
+import org.xbill.DNS.Name;
+
+/** A base class for all types of things that might fail when making a DNS lookup. */
+public class LookupFailedException extends RuntimeException {
+  private final Name name;
+  private final int type;
+
+  public LookupFailedException() {
+    super();
+    name = null;
+    type = 0;
+  }
+
+  public LookupFailedException(String message) {
+    super(message);
+    name = null;
+    type = 0;
+  }
+
+  /**
+   * Construct a LookupFailedException that also specifies the name and type of the lookup that
+   * failed.
+   *
+   * @param name the name that caused the failure.
+   * @param type the type that caused the failure.
+   */
+  public LookupFailedException(Name name, int type) {
+    this.name = name;
+    this.type = type;
+  }
+}
--- a/src/main/java/org/xbill/DNS/lookup/NoSuchDomainException.java
+++ b/src/main/java/org/xbill/DNS/lookup/NoSuchDomainException.java
@@ -1,0 +1,13 @@
+package org.xbill.DNS.lookup;
+
+import org.xbill.DNS.Name;
+
+/**
+ * Thrown to indicate that no data is associated with the given name, as indicated by the NXDOMAIN
+ * response code as specified in RF2136 Section 2.2.
+ */
+public class NoSuchDomainException extends LookupFailedException {
+  public NoSuchDomainException(Name name, int type) {
+    super(name, type);
+  }
+}
--- a/src/main/java/org/xbill/DNS/lookup/NoSuchRRSetException.java
+++ b/src/main/java/org/xbill/DNS/lookup/NoSuchRRSetException.java
@@ -1,0 +1,13 @@
+package org.xbill.DNS.lookup;
+
+import org.xbill.DNS.Name;
+
+/**
+ * Thrown to indicate that records of the name and type queried does not exist, corresponding to the
+ * NXRRSET return code as specified in RFC2136 Section 2.2.
+ */
+public class NoSuchRRSetException extends LookupFailedException {
+  public NoSuchRRSetException(Name name, int type) {
+    super(name, type);
+  }
+}
--- a/src/main/java/org/xbill/DNS/lookup/ServerFailedException.java
+++ b/src/main/java/org/xbill/DNS/lookup/ServerFailedException.java
@@ -1,0 +1,8 @@
+// SPDX-License-Identifier: BSD-2-Clause
+package org.xbill.DNS.lookup;
+
+/**
+ * Represents a server failure, that the upstream server responding to the request returned a
+ * SERVFAIL status.
+ */
+public class ServerFailedException extends LookupFailedException {}
--- a/src/test/java/org/xbill/DNS/LookupTest.java
+++ b/src/test/java/org/xbill/DNS/LookupTest.java
@@ -1,6 +1,9 @@
 // SPDX-License-Identifier: BSD-2-Clause
 package org.xbill.DNS;
 
+import static java.util.Arrays.asList;
+import static java.util.Collections.singletonList;
+import static java.util.stream.Collectors.joining;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertNull;
 import static org.junit.jupiter.api.Assertions.assertThrows;
@@ -15,7 +18,6 @@ import java.io.InterruptedIOException;
 import java.net.InetAddress;
 import java.util.List;
 import java.util.function.Function;
-import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeEach;
@@ -24,7 +26,10 @@ import org.mockito.ArgumentCaptor;
 import org.mockito.Mockito;
 
 public class LookupTest {
-  public static final Name DUMMY_NAME = Name.fromConstantString("ignored.");
+  public static final Name DUMMY_NAME = Name.fromConstantString("to.be.replaced.");
+  public static final String LONG_LABEL =
+      IntStream.range(0, 63).mapToObj(i -> "a").collect(joining());
+
   private Resolver mockResolver;
 
   @BeforeEach
@@ -128,6 +133,10 @@ public class LookupTest {
         queries.get(1).getQuestion());
 
     assertEquals(1, results.length);
+    assertEquals(Name.fromConstantString("target.example.com."), results[0].getName());
+    assertEquals(
+        singletonList(Name.fromConstantString("hostX.first.example.com.")),
+        asList(lookup.getAliases()));
   }
 
   Message maybeCnameAnswer(Message query) {
@@ -140,6 +149,45 @@ public class LookupTest {
           }
           return new ARecord(DUMMY_NAME, DClass.IN, 0, InetAddress.getLoopbackAddress());
         });
+  }
+
+  @Test
+  void testRun_firstSearchPathHitsServFail() throws Exception {
+    wireUpMockResolver(mockResolver, this::firstSERVFAILThenA);
+
+    Lookup lookup = makeLookupWithResolver(mockResolver, "host2");
+    lookup.setSearchPath("first.example.com", "second.example.com");
+    Record[] results = lookup.run();
+
+    ArgumentCaptor<Message> messageCaptor = ArgumentCaptor.forClass(Message.class);
+    verify(mockResolver, times(2)).send(messageCaptor.capture());
+
+    List<Message> queries = messageCaptor.getAllValues();
+
+    assertEquals(
+        Record.newRecord(
+            Name.fromConstantString("host2.first.example.com."), Type.A, DClass.IN, 0L),
+        queries.get(0).getQuestion());
+    assertEquals(
+        Record.newRecord(
+            Name.fromConstantString("host2.second.example.com."), Type.A, DClass.IN, 0L),
+        queries.get(1).getQuestion());
+
+    assertEquals(1, results.length);
+    assertEquals(Lookup.SUCCESSFUL, lookup.getResult());
+  }
+
+  Message firstSERVFAILThenA(Message query) {
+    Message answer = new Message(query.getHeader().getID());
+    answer.addRecord(query.getQuestion(), Section.QUESTION);
+    Name questionName = query.getQuestion().getName();
+    if (questionName.equals(Name.fromConstantString("host2.first.example.com."))) {
+      answer.getHeader().setRcode(Rcode.SERVFAIL);
+    } else {
+      Record r = new ARecord(questionName, DClass.IN, 0, InetAddress.getLoopbackAddress());
+      answer.addRecord(r, Section.ANSWER);
+    }
+    return answer;
   }
 
   @Test
@@ -243,11 +291,9 @@ public class LookupTest {
   void testRun_concatenatedNameTooLong() throws Exception {
     wireUpMockResolver(mockResolver, this::simpleAnswer);
 
-    String longName = IntStream.range(0, 63).mapToObj(i -> "a").collect(Collectors.joining());
-
-    Lookup lookup = makeLookupWithResolver(mockResolver, longName);
+    Lookup lookup = makeLookupWithResolver(mockResolver, LONG_LABEL);
     // search path has a suffix that will make the combined name too long
-    lookup.setSearchPath(String.format("%s.%s.%s", longName, longName, longName));
+    lookup.setSearchPath(String.format("%s.%s.%s", LONG_LABEL, LONG_LABEL, LONG_LABEL));
     Record[] results = lookup.run();
 
     ArgumentCaptor<Message> messageCaptor = ArgumentCaptor.forClass(Message.class);
@@ -257,7 +303,7 @@ public class LookupTest {
     // ignored, and resolution falls back to converting longName to an absolute name and querying
     // that
     assertEquals(
-        Record.newRecord(Name.fromConstantString(longName + "."), Type.A, DClass.IN, 0L),
+        Record.newRecord(Name.fromConstantString(LONG_LABEL + "."), Type.A, DClass.IN, 0L),
         messageCaptor.getValue().getSection(Section.QUESTION).get(0));
 
     assertEquals(1, results.length);
@@ -287,7 +333,7 @@ public class LookupTest {
   }
 
   @Test
-  void testLookup_constructorFailsWithMetaTypes() throws TextParseException {
+  void testLookup_constructorFailsWithMetaTypes() {
     assertThrows(IllegalArgumentException.class, () -> new Lookup("example.com.", Type.OPT));
   }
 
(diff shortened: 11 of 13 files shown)


ibauersachs:
> I have started experimenting with an asynchronous lookup mechanism. Please note that this is a draft and not not feature complete. I just want to give the opportunity to gather some early feedback. In particular I am curious to to hear about:
> 
> * The name `DNSSession` for the new long lived API endpoint. Session is kind of vague on purpose, to allow it to have other public methods in the future that interact with DNS servers. I want to avoid Session because of the inconvenience of having it clash with for example the Session classes in the Neo4j or Cassandra.

The prefix is goog, but please try to avoid the uppercasing for new stuff, i.e. use `DnsSession`. I don't have other thoughts yet.

> * What do we think about putting an exception hierarchy in a separate module? My main reason for going there is that the org.xbill.DNS module is pretty crowded, but this is a weak preference.

A separate package for all things lookup, e.g. org.xbill.DNS.lookup is fine, but placing exceptions in a dedicated package is IMO weird. `IOException` for example is in `java.io`, not `java.io.exceptions`.

src/main/java/org/xbill/DNS/DNSSession.java
```diff
@@ -0,0 +1,129 @@
+package org.xbill.DNS;
```
ibauersachs, line 1: Please add `// SPDX-License-Identifier: BSD-2-Clause` to new files to clearly indicate their license.

src/main/java/org/xbill/DNS/DNSSession.java
```diff
@@ -0,0 +1,129 @@
...
+   */
+  public void setMaxRedirects(int maxRedirects) {
+    this.maxRedirects = maxRedirects;
+  }
+
+  /**
+   * Make an asynchronous lookup of the provided name.
+   *
+   * @param name the name to look up.
+   * @param type the type to look up, values should correspond to constants in Type.
```
ibauersachs, line 47: Link to classes with `{@link Type}` (in other places too)

src/main/java/org/xbill/DNS/DNSSession.java
```diff
@@ -0,0 +1,129 @@
...
+      LookupResult result = dealWithEmptyAnswer(allAnswers, response.getHeader().getRcode());
+      if (result != null) {
+        future.complete(result);
+        return;
+      }
+      Record firstAnswer = allAnswers.get(0);
+
+      if (firstAnswer.getType() == Type.DNAME || firstAnswer.getType() == Type.CNAME) {
+        resolver
+            .sendAsync(Message.newQuery(buildRedirectQuery(response)))
```
ibauersachs, line 82: I know this is like the original Lookup class. But is it even legal for a non-recursive resolver to follow CNAMEs if the answer from a recursing resolver is incomplete?

src/main/java/org/xbill/DNS/LookupResult.java
```diff
@@ -0,0 +1,61 @@
...
+  public List<Record> get() {
+    return records;
+  }
+
+  public AdditionalDetail getAdditionalDetail() {
+    return additionalDetail;
+  }
+
+  @Override
+  public boolean equals(Object o) {
```
ibauersachs, line 45: You can use [Lombok](https://projectlombok.org/features/EqualsAndHashCode) for this: add `@EqualsAndHashCode` to the class.

src/main/java/org/xbill/DNS/LookupResult.java
```diff
@@ -0,0 +1,61 @@
...
+    return records.equals(that.records) && additionalDetail == that.additionalDetail;
+  }
+
+  @Override
+  public int hashCode() {
+    return hash(records, additionalDetail);
+  }
+
+  @Override
+  public String toString() {
```
ibauersachs, line 58: Can also be done with [Lombok](https://projectlombok.org/features/ToString) `@ToString`.

src/test/java/org/xbill/DNS/DNSSessionTest.java
```diff
@@ -0,0 +1,225 @@
...
+  }
+
+  @Test
+  public void lookupAsync_NXDOMAIN() throws Exception {
+    wireUpMockResolver(mockResolver, q -> fail(q, Rcode.NXDOMAIN));
+
+    DNSSession dnsSession = new DNSSession(mockResolver);
+    CompletionStage<LookupResult> resultFuture = dnsSession.lookupAsync(name("a.b."), A, IN);
+
+    assertEquals(new LookupResult(emptyList(), NXDOMAIN), resultFuture.toCompletableFuture().get());
```
ibauersachs, line 118: I'm just not convinced that an empy list is really the best answer for NXDOMAIN, but neither are `Optional`s or exceptions.

Ideally the return type would be tuple of (result, records) so callers would be forced to not ignore it.

src/main/java/org/xbill/DNS/LookupResult.java
```diff
@@ -0,0 +1,61 @@
...
+
+import static java.util.Objects.hash;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import org.xbill.DNS.exceptions.AdditionalDetail;
+
+/** LookupResult instances holds the result of a successful lookup operation. */
+public class LookupResult {
```
ibauersachs, line 11: If the intention is that this class is immutable, make it `final`.

src/main/java/org/xbill/DNS/DNSSession.java
```diff
@@ -0,0 +1,129 @@
...
+    if (answer.isEmpty()) {
+      switch (rcode) {
+        case Rcode.NXDOMAIN:
+          return new LookupResult(answer, AdditionalDetail.NXDOMAIN);
+        case Rcode.NXRRSET:
+          return new LookupResult(answer, AdditionalDetail.NXRRSET);
+        case Rcode.SERVFAIL:
+          throw new ServerFailedException();
+        default:
+          throw new LookupFailedException(format("Unknown non-success error code %d", rcode));
```
ibauersachs, line 124: Use `Rcode.string(rcode)` to translate the numeric value. The underlying mnemonic class takes care of handling unknown values, so it's guaranteed to have a sensible message.

src/test/java/org/xbill/DNS/DNSSessionTest.java
ibauersachs: Do you have an idea/examples of how non-unit test sync and async code should look like?

src/main/java/org/xbill/DNS/DNSSession.java
```diff
@@ -0,0 +1,129 @@
...
+
+/**
+ * DNSSession provides facilities to make DNS Queries. A DNSSession is intended to be long lived,
+ * and it's behaviour can be modified by calling the setter methods.
+ */
+public class DNSSession {
+  private static final int DEFAULT_MAX_ITERATIONS = 16;
+
+  private final Resolver resolver;
+  private volatile int maxRedirects = DEFAULT_MAX_ITERATIONS;
```
ibauersachs, line 22: Why volatile?

src/main/java/org/xbill/DNS/DNSSession.java
```diff
@@ -0,0 +1,129 @@
...
+import org.xbill.DNS.exceptions.InvalidZoneDataException;
+import org.xbill.DNS.exceptions.LookupFailedException;
+import org.xbill.DNS.exceptions.RedirectOverflowException;
+import org.xbill.DNS.exceptions.ServerFailedException;
+
+/**
+ * DNSSession provides facilities to make DNS Queries. A DNSSession is intended to be long lived,
+ * and it's behaviour can be modified by calling the setter methods.
+ */
+public class DNSSession {
```
ibauersachs, line 18: - Session somewhat implies it needs closing or that it expires, not sure this really applies here
- The async stuff creates some interesting threading issues once there are more config options than a simple integer. E.g. what if there's a cache setter, and the cache is changed while an async lookup is still in progress?
- The existing Cache class is not thread safe (IIRC), so it cannot be reused

src/main/java/org/xbill/DNS/DNSSession.java
nresare: Adding..

src/main/java/org/xbill/DNS/DNSSession.java
nresare: I agree. It is strange. I just wanted to make it a drop in replacement. Come to think of it I think that it makes sense to have maxRedirects default to 0 and have that mean to not follow redirects and have Lookup use the old behaviour at least until the next major release.

src/main/java/org/xbill/DNS/LookupResult.java
nresare: Thank you for reminding me. I have encountered lombok at the margins but never really engaged with it. This seems very useful. It seems if you add `lombok.addLombokGeneratedAnnotation = true` to a `lombok.config` in the toplevel directory it will exclude generated code from the code coverage calculation as well. I'll add that.

src/main/java/org/xbill/DNS/LookupResult.java
nresare: changing

src/test/java/org/xbill/DNS/DNSSessionTest.java
nresare: Let me think some more about how calling code might look and get back to you on that.

src/main/java/org/xbill/DNS/LookupResult.java
nresare: Good catch, updated

src/main/java/org/xbill/DNS/DNSSession.java
nresare: That is great. At some point I think we should discuss moving to proper enums for the mnemonic classes, but that is for a later time.

src/main/java/org/xbill/DNS/DNSSession.java
nresare: It will ensure that reads after a write will get the new value even if it is in a different thread. So, if thread A calls `setMaxRedirects()` with a new value and then thread B calls `lookupAsync()` the updated value will be visible to thread B. Without volatile, thread B might hold a cached copy of the old value indefinitely and not ever see the update.

src/main/java/org/xbill/DNS/DNSSession.java
ibauersachs: maxRedirects in the existing Lookup class uses that parameter also to follow the chain within a Message, which is fine. Not following CNAMEs by sending more queries to a (recursive) resolvers is something that can be done in a minor version, or changed via an option.

nresare:
A separate lookup package seems like a good idea to me. I'm happy to go with `DnsSession` for now, looking at the list of classes it is easy to get the impression that we do appreviations in all caps, but I guess that is just because the RFCs uses that for the different data types.

src/main/java/org/xbill/DNS/DNSSession.java
nresare: I like immutable classes, with the drawback that configuration typically would be through a long set of constructor args or via the Builder pattern which is a bit verbose. Let me think some more about this

src/main/java/org/xbill/DNS/DNSSession.java
nresare: Sure. How do we feel about adding copyright statements? 

I guess the main purpose for such things would be if someone would make an effort to contact all copyright holders to investigate the viability of a license change in the future. I am happy with or without, and I'm happy with 2 clause BSD.

src/main/java/org/xbill/DNS/LookupResult.java
ibauersachs: I don't think I can ever code Java without Lombok again :-)

Please make sure the option doesn't introduce a runtime dependency on Lombok via the `@lombok.Generated` annotation.

src/main/java/org/xbill/DNS/DNSSession.java
ibauersachs: Enums: yes and no. A lot of stuff probably should stay as it is, not just because of the API compatibility, but also because DNS is extensible and with the integers everything still works when new stuff is introduced even if we don't update anything.

ibauersachs:
> A separate lookup package seems like a good idea to me. I'm happy to go with `DnsSession` for now, looking at the list of classes it is easy to get the impression that we do appreviations in all caps, but I guess that is just because the RFCs uses that for the different data types.

The all-caps is mostly historical and I would e.g. create `DnssecRecord` instead of `DNSSECRecord` today. Except for the Record implementations, I'd go for the more sane version. But I won't insist on it if you prefer `DNSSession`.

src/main/java/org/xbill/DNS/DNSSession.java
ibauersachs: A builder seems appropriate - and Lombok to the rescue ;-)

src/main/java/org/xbill/DNS/LookupResult.java
nresare: Some googling seems to suggest that annotations that byte code parser encountering annotations that are not on the classpath will ignore them, at least after this bug was fixed: https://bugs.java.com/bugdatabase/view_bug.do?bug_id=6322301

src/main/java/org/xbill/DNS/DNSSession.java
ibauersachs: The Git history is enough to follow the copyright. The Kernel and systemd do it the same with the simple SPDX comment.

src/main/java/org/xbill/DNS/LookupResult.java
ibauersachs: Guava has a bunch of transitive dependencies from annotation libraries that shouldn't be necessary. Not quite sure where the issue lies exactly, Maven, Java, an IDE, whatever, but they need to be there that stuff doesn't fail. See e.g. the discussions in google/guava#3302.

ibauersachs:
About that "session", yesterday I tought about using "LookupSession". DnsSession might be a bit too generic. Just an idea.

src/main/java/org/xbill/DNS/lookup/DnsSession.java
```diff
@@ -0,0 +1,214 @@
...
+            searchPath.stream()
+                .map(
+                    name -> {
+                      try {
+                        return Name.concatenate(name, Name.root);
+                      } catch (NameTooLongException e) {
+                        throw new IllegalArgumentException("Search path name too long");
+                      }
+                    })
+                .collect(Collectors.toCollection(ArrayList::new));
```
ibauersachs, line 70: Use `Collectors.toList()`

src/main/java/org/xbill/DNS/lookup/DnsSession.java
```diff
@@ -0,0 +1,214 @@
...
+   * never empty, as it will at the very least always contain name.
+   */
+  Stream<Name> expandName(Name name) {
+    if (name.isAbsolute()) {
+      return Stream.of(name);
+    }
+    if (name.length() == 255) {
+      // handle this special case where the name is relative but can not be made absolute as it
+      // is at it's maximum length and nothing can be appended.
+      throw new IllegalArgumentException("Name is too long");
```
ibauersachs, line 102: Can't you use `NameTooLongException`?

src/main/java/org/xbill/DNS/DNSSession.java
nresare: I'll look into changing that as a followup to this, but I feel I would like to land this first and implementing the old behaviour first seems like a good idea.

src/main/java/org/xbill/DNS/LookupResult.java
nresare: Having taken a brief look at the referenced issues I don't think that these are comparable, but I will verify this locally.

src/test/java/org/xbill/DNS/DNSSessionTest.java
nresare: I'll write something as a new comment below, as this is a comment on an outdated piece of code.

src/main/java/org/xbill/DNS/DNSSession.java
nresare: Good point!

src/main/java/org/xbill/DNS/DNSSession.java
nresare: Since we have moved to a builder pattern now, this point is moot.

src/main/java/org/xbill/DNS/lookup/DnsSession.java
nresare: Lombook generates the searchPath property as being of type ArrayList, so the code that makes the names absolute needs to return an ArrayList as well to be able to update the value.

src/main/java/org/xbill/DNS/lookup/DnsSession.java
nresare: Since `NameTooLongException` is a checked exception it would add an exception to the `lookupAsync` and one for a pretty esoteric case as well. (Since it takes a Name and names can't be constructed if they are too long, it is only for the case where it is a maximally long relative name that can not be made absolute).

I think this is a good use case for a checked exception, as it specifies exactly what is going on for the edge case but for the common case it can be ignored and the failure will bubble up the call stack. Callers that want to deal with it can do that, but callers that don't doesn't need to.

Moreover, since I envision lookupAsync() to be used in chains of CompletionStage based async code, having it throw a checked exception would make those chains a lot more verbose as it would need to be either handled out of the call stack or wrapped in an unchecked exception.

nresare: Come to think of it I think that the better way to deal with this is to change Name to fail creation for relative names of length 255. That way the caller would need to handle that the same way as it would need to handle other Name creations that are much longer.

nresare:
I'm happy to change the name to LookupSession. If we would want to mix lookups and other operations with resolvers (zone transfers and updates comes to mind) in the future we can always create a new interface that has all of those aspects baked in.

nresare:
After having added a few more bits and pieces, I believe that the code is feature complete now. Things left to do:

1. Some more analysis on the pros and cons of the different ways to returning empty non-exceptional results
2. Get rid of the search path too long code path.
3. Add a delombok stage to the build and have javadoc be generated off of that, as the `@Builder` related methods in LookupSession is pretty important to it's functioning

src/main/java/org/xbill/DNS/lookup/DnsSession.java
ibauersachs: Ah. But then `.forEach(searchPath::add);` is IMO more better than overwriting an already instantiated field.

src/main/java/org/xbill/DNS/lookup/DnsSession.java
ibauersachs: Without following the code too much now, what happens with two search paths where only the second will generate a NTLE? If that is communicated as a failed future, you could do the same with with the first path (by try/catching expandName at the call site and immediately returning a failed future. No need for a checked exception in the signature of lookupAsync.

As for limiting the Name length, not sure. Let's discuss this somwhere else.

src/main/java/org/xbill/DNS/LookupResult.java
ibauersachs: I may have picked up the wrong issues in Guava, but all the stuff in their dependency chain that comes along with Guava (failureaccess, listenablefuture, jsr305, errorprone, checker-qual, etc.) is all just for static analysis tools but somehow needed at runtime, at least in some situations (that aren't obvious).

ibauersachs:
> I'm happy to change the name to LookupSession.

I don't know if it's really better, just an idea I had while cooking :-)

> If we would want to mix lookups and other operations with resolvers (zone transfers and updates comes to mind) in the future we can always create a new interface that has all of those aspects baked in.

Not sure how you inted to mix that. In any case, I don't see how a zone transfer/update could be logically mixed with the options set in the builder (ndots, search path, cache). I mean okay, you could try to update a relative name, but that just makes no sense at all (and the constructor of Update forbids it).

src/main/java/org/xbill/DNS/lookup/DnsSession.java
nresare: Ah, I did not think of that. Having any checked exception be passed on through the CompletionStage is definitely one way of doing it.

I'm okay with that, although I think that changing the Name constructor to disallow it is the better tradeoff, but we can discuss that in the PR that I just opened. 

My fairly strong opinion is that checked exceptions as designed in Java is a flawed concept, but that is not something that we will fix tonight.

nresare:
> > I'm happy to change the name to LookupSession.
> 
> I don't know if it's really better, just an idea I had while cooking :-)
> 

Having now seen both in front of me, I think that `LookupSession` is a better name.

nresare:
So, signalling NXDOMAIN/NXRRSET as a lookup result. I see 5 options in the case of a query resulting in a NXDOMAIN/NXRRSET response from the resolver:

1. The current variant. LookupResult.getResult() returning an empty list.
2. having LookupResult.getResult() return null on NXDOMAIN/NXRRSET.
3. Throwing an exception
4. Having getResult() return an `Optional<List<Record>>` that is empty 
5. have getResult() throw an exception when called

My read is that we would like calling code dealing with this case, and we we would like to nudge callers to take it into account. I think 3-5 does that. 3 has the drawback of using exceptions for a decidedly non-exceptional activity. Take for example code that would make lookups in various DNS based junk email blacklists. Most responses will be NXDOMAIN and although creating Exceptions is not as expensive that that it once was, it still has a cost.

I think that the method signature in 4 is bit too long to look good, but it has the benefit of it being syntactically straightforward to promote the case into an exception (which I assume many callers would want) with `result.getResult().orElseThrow()`.

Come to think of it, if we went with 2, consumers could get that pattern if they wanted with `Optional.of(result.getResult()).orElseThrow()`.

5 has some appeal, as the "high performance" user could look at the additionalDetail field and avoid generating an exception and the casual user that just want NXDOMAIN to be treated as any other failure somewhere up the call stack could get that. However, I am skeptical of diverging from the principle of least astonishment to have a getter throw an exception.

I am leaning towards 2, as it gives a clean way to have Optional.orElseThrow() to get the exception and I can't really think of a case where the empty list would be easier to consume from than handling the null value.

nresare:
I rebased and squashed the commits and removed the draft flag, as I think that we are mostly in agreement about this change. Please let me know your thoughts on the NXDOMAIN options I outline above.

nresare:
I have now verified that using an artefact built with the lombok annotated byte code (to prevent code coverage from measuring generated bytecode) I can run a project with no lombok dependencies, only dnsjava and there are no issues loading the byte code from that jar. 

```
[INFO] com.resare.maven-test:maven-test:jar:0.0.1
[INFO] +- org.junit.jupiter:junit-jupiter-api:jar:5.7.0:test
[INFO] |  +- org.apiguardian:apiguardian-api:jar:1.1.0:test
[INFO] |  +- org.opentest4j:opentest4j:jar:1.2.0:test
[INFO] |  \- org.junit.platform:junit-platform-commons:jar:1.7.0:test
[INFO] +- org.junit.jupiter:junit-jupiter-engine:jar:5.7.0:test
[INFO] |  \- org.junit.platform:junit-platform-engine:jar:1.7.0:test
[INFO] \- dnsjava:dnsjava:jar:3.4.0-SNAPSOT:compile
[INFO]    \- org.slf4j:slf4j-api:jar:1.7.30:compile
```

src/main/java/org/xbill/DNS/lookup/DnsSession.java
ibauersachs: The concept is broken, yes. Especially now with the lambdas. But Java has been born disabled. Signed bytes. Sigh.

ibauersachs:
> So, signalling NXDOMAIN/NXRRSET as a lookup result. I see 5 options in the case of a query resulting in a NXDOMAIN/NXRRSET response from the resolver:
> 
> 1. The current variant. LookupResult.getResult() returning an empty list.

I checked out this branch to code an example (based on some code I had lying around) and immediately fell into the trap of expecting an answer because I used the name before, but it was now NXDOMAIN. The empy list thus was surprising.

> 2. having LookupResult.getResult() return null on NXDOMAIN/NXRRSET.

Nah, null for Lists is unexpected.

> 3. Throwing an exception

I don't think this is so bad. In this flow I defintely expect to have a valid, usable answer or an exception, not something in between:

```java
    SimpleResolver sr = new SimpleResolver("8.8.8.8");
    LookupSession l = LookupSession.builder().resolver(sr).ndots(1).maxRedirects(16).build();
    l.lookupAsync(Name.fromConstantString("macmini.local"), Type.A, DClass.IN)
        .toCompletableFuture()
        .thenAccept(
            res -> {
              for (Record r : res.getRecords()) {
                System.out.println(r);
              }
            })
        .exceptionally(
            e -> {
              System.out.println(e.getMessage());
              return null;
            })
        .get();
```

> 4. Having getResult() return an `Optional<List<Record>>` that is empty

Maybe, not sure.

> 5. have getResult() throw an exception when called

Definitely not an option.

> 
> My read is that we would like calling code dealing with this case, and we we would like to nudge callers to take it into account. I think 3-5 does that. 3 has the drawback of using exceptions for a decidedly non-exceptional activity. Take for example code that would make lookups in various DNS based junk email blacklists.

FWIW, the blacklists always return an answer, the status is encoded in the last octet of a 127.0.0.0/8 address.

> Most responses will be NXDOMAIN and although creating Exceptions is not as expensive that that it once was, it still has a cost.

It's bad style to rely on exceptions for expected outcomes. BUT in an async API I'm not so sure anymore. Plus the exception isn't really thrown anymore, so it's less of overhead than an actually thrown exception (but still more than enum).

> I think that the method signature in 4 is bit too long to look good, but it has the benefit of it being syntactically straightforward to promote the case into an exception (which I assume many callers would want) with `result.getResult().orElseThrow()`.

Something in between could be an option (configurable via builder) to let the consumer decide if NX* should result in a failed or succeeded future. Not sure if it's a good idea either.

> 
> Come to think of it, if we went with 2, consumers could get that pattern if they wanted with `Optional.of(result.getResult()).orElseThrow()`.
> 
> 5 has some appeal, as the "high performance" user could look at the additionalDetail field and avoid generating an exception and the casual user that just want NXDOMAIN to be treated as any other failure somewhere up the call stack could get that. However, I am skeptical of diverging from the principle of least astonishment to have a getter throw an exception.
> 
> I am leaning towards 2, as it gives a clean way to have Optional.orElseThrow() to get the exception and I can't really think of a case where the empty list would be easier to consume from than handling the null value.

I haven't seen Optionals used like this.

ibauersachs:
> I have now verified that using an artefact built with the lombok annotated byte code (to prevent code coverage from measuring generated bytecode) I can run a project with no lombok dependencies, only dnsjava and there are no issues loading the byte code from that jar.
> 
> ```
> [INFO] com.resare.maven-test:maven-test:jar:0.0.1
> [INFO] +- org.junit.jupiter:junit-jupiter-api:jar:5.7.0:test
> [INFO] |  +- org.apiguardian:apiguardian-api:jar:1.1.0:test
> [INFO] |  +- org.opentest4j:opentest4j:jar:1.2.0:test
> [INFO] |  \- org.junit.platform:junit-platform-commons:jar:1.7.0:test
> [INFO] +- org.junit.jupiter:junit-jupiter-engine:jar:5.7.0:test
> [INFO] |  \- org.junit.platform:junit-platform-engine:jar:1.7.0:test
> [INFO] \- dnsjava:dnsjava:jar:3.4.0-SNAPSOT:compile
> [INFO]    \- org.slf4j:slf4j-api:jar:1.7.30:compile
> ```

Okay, but the annotations need to go on the first sign of trouble.

src/main/java/org/xbill/DNS/lookup/AdditionalDetail.java
```diff
@@ -0,0 +1,15 @@
+// SPDX-License-Identifier: BSD-2-Clause
+package org.xbill.DNS.lookup;
+
+public enum AdditionalDetail {
```
ibauersachs, line 4: - Missing class-level Javadoc
- Needs a SUCESSFUL (or OK) case
- A better name might be `LookupResultDetail`

src/main/java/org/xbill/DNS/lookup/InvalidZoneDataException.java
```diff
@@ -0,0 +1,8 @@
+// SPDX-License-Identifier: BSD-2-Clause
+package org.xbill.DNS.lookup;
+
+public class InvalidZoneDataException extends LookupFailedException {
```
ibauersachs, line 4: class-level Javadoc

src/main/java/org/xbill/DNS/lookup/LookupResult.java
```diff
@@ -0,0 +1,40 @@
...
+import org.xbill.DNS.Name;
+import org.xbill.DNS.Record;
+
+/** LookupResult instances holds the result of a successful lookup operation. */
+@ToString
+@EqualsAndHashCode
+public final class LookupResult {
+  /** An unmodifiable list of records that this instance wraps, may not be null but can be empty */
+  @Getter private final List<Record> records;
+  /** provides additional detail when records is empty */
```
ibauersachs, line 19: - Capitalize the beginning
- Use {@link #getRecords()} to refer to the list

src/main/java/org/xbill/DNS/lookup/LookupResult.java
```diff
@@ -0,0 +1,40 @@
...
+  @Getter private final AdditionalDetail additionalDetail;
+  /** tracks the different CNAME and DNAME aliases for this name */
```
ibauersachs, line 21: Capitalize

src/main/java/org/xbill/DNS/lookup/LookupResult.java
```diff
@@ -0,0 +1,40 @@
...
+  @Getter private List<Name> aliases;
+
+  /**
+   * Construct an instance with the provided records.
+   *
+   * @param records a list of records to return, or null if there was no response
+   * @param additionalDetail additional detail on this response, such as the reason for records
+   *     being null.
+   */
+  public LookupResult(List<Record> records, AdditionalDetail additionalDetail, List<Name> aliases) {
```
ibauersachs, line 31: I'd keep this package-private as long as there's no need to have it public

src/main/java/org/xbill/DNS/lookup/LookupResult.java
```diff
@@ -0,0 +1,40 @@
...
+    this.records = Collections.unmodifiableList(new ArrayList<>(records));
+    this.aliases = aliases;
+    if (records.isEmpty() && additionalDetail == null) {
+      throw new IllegalArgumentException(
+          "Empty records can not be combined with null additionalDetail");
+    }
+    this.additionalDetail = additionalDetail;
```
ibauersachs, line 38: Guarantee that this is not null with the OK status (as Java cannot (yet) handle null in switch statements).

src/main/java/org/xbill/DNS/lookup/LookupSession.java
```diff
@@ -0,0 +1,299 @@
...
+import org.xbill.DNS.NameTooLongException;
+import org.xbill.DNS.Rcode;
+import org.xbill.DNS.Record;
+import org.xbill.DNS.Resolver;
+import org.xbill.DNS.Section;
+import org.xbill.DNS.SetResponse;
+import org.xbill.DNS.Type;
+
+/**
+ * DNSSession provides facilities to make DNS Queries. A DNSSession is intended to be long lived,
```
ibauersachs, line 43: Old name used, refers to setters instead of builder

src/main/java/org/xbill/DNS/lookup/LookupSession.java
```diff
@@ -0,0 +1,299 @@
...
+ */
+@Builder(buildMethodName = "innerBuild")
+public class LookupSession {
+  private static final int DEFAULT_MAX_ITERATIONS = 16;
+  private static final int DEFAULT_NDOTS = 1;
+  /** The {@link Resolver} to use to look up records. */
+  @NonNull private final Resolver resolver;
+  /**
+   * Sets the maximum number of CNAME or DNAME redirects allowed before lookups with fail with
+   * {@link RedirectOverflowException}.
```
ibauersachs, line 54: `... Defaults to {@value #DEFAULT_MAX_ITERATIONS}.`

src/main/java/org/xbill/DNS/lookup/LookupSession.java
```diff
@@ -0,0 +1,299 @@
...
+import org.xbill.DNS.SetResponse;
+import org.xbill.DNS.Type;
+
+/**
+ * DNSSession provides facilities to make DNS Queries. A DNSSession is intended to be long lived,
+ * and it's behaviour can be modified by calling the setter methods.
+ */
+@Builder(buildMethodName = "innerBuild")
+public class LookupSession {
+  private static final int DEFAULT_MAX_ITERATIONS = 16;
```
ibauersachs, line 48: Might be worth it having it public. Not sure.

src/main/java/org/xbill/DNS/lookup/LookupSession.java
```diff
@@ -0,0 +1,299 @@
...
+  /** The {@link Resolver} to use to look up records. */
+  @NonNull private final Resolver resolver;
+  /**
+   * Sets the maximum number of CNAME or DNAME redirects allowed before lookups with fail with
+   * {@link RedirectOverflowException}.
+   */
+  @Builder.Default private final int maxRedirects = DEFAULT_MAX_ITERATIONS;
+  /**
+   * Sets the threshold for the number of dots which much appear in a name before it is considered
+   * absolute. The default is {@code 1}, meaning meaning that if there are any dots in a name, the
```
ibauersachs, line 59: `{@value #NDOTS}`

Should this default constant be shared with the default in `org.xbill.DNS.config.ResolverConfigProvider`? And I've forgotten how this default interacts with the one from the ResolverConfig anyway.

src/main/java/org/xbill/DNS/lookup/LookupSession.java
```diff
@@ -0,0 +1,299 @@
...
+                    name -> {
+                      try {
+                        return Name.concatenate(name, Name.root);
+                      } catch (NameTooLongException e) {
+                        throw new IllegalArgumentException("Search path name too long");
+                      }
+                    })
+                .collect(Collectors.toCollection(ArrayList::new));
+      }
+      return innerBuild();
```
ibauersachs, line 98: The innerBuild method is public. It needs to be at least package-private (found this: https://rusyasoft.github.io/java,%20lombok/2020/02/14/Overriding-build-setter-lombok-builder/, but maybe there's a better way).

src/main/java/org/xbill/DNS/lookup/LookupSession.java
```diff
@@ -0,0 +1,299 @@
...
+
+  /**
+   * Make an asynchronous lookup of the provided name.
+   *
+   * @param name the name to look up.
+   * @param type the type to look up, values should correspond to constants in {@link Type}.
+   * @param dclass the class to look up, values should correspond to constants in {@link DClass}.
+   * @return A {@link CompletionStage} what will yield the eventual lookup result.
+   */
+  public CompletionStage<LookupResult> lookupAsync(Name name, int type, int dclass) {
```
ibauersachs, line 110: add an overload without DClass that defaults to DClass.IN, and maybe one that takes a String instead of a Name

src/main/java/org/xbill/DNS/lookup/LookupSession.java
```diff
@@ -0,0 +1,299 @@
...
+  /**
+   * Generate a stream of names according to the search path application semantics. The semantics of
+   * this is a bit odd, but they are inherited from Lookup.java. Note that the stream returned is
+   * never empty, as it will at the very least always contain name.
+   */
+  Stream<Name> expandName(Name name) throws NameTooLongException {
+    if (name.isAbsolute()) {
+      return Stream.of(name);
+    }
+    if (name.length() == 255) {
```
ibauersachs, line 133: This is superfluos with #165, right?

src/main/java/org/xbill/DNS/lookup/LookupSession.java
```diff
@@ -0,0 +1,299 @@
...
+   * If set to true, cached results with multiple records will be returned with the starting point
+   * shifted one step per request.
+   */
+  @Builder.Default private final boolean cycleResults = false;
+
+  /**
+   * Configures the Cache instances to be used for lookups for the different {@link DClass} values.
+   */
+  @Singular("cache")
+  private final Map<Integer, Cache> caches;
```
ibauersachs, line 78: Is there a default Cache as before? How can I set one without bothering about what a DClass is?

src/main/java/org/xbill/DNS/lookup/LookupSession.java
```diff
@@ -0,0 +1,299 @@
...
+  @Builder.Default private final int maxRedirects = DEFAULT_MAX_ITERATIONS;
+  /**
+   * Sets the threshold for the number of dots which much appear in a name before it is considered
+   * absolute. The default is {@code 1}, meaning meaning that if there are any dots in a name, the
+   * name will be tried first as an absolute name.
+   */
+  @Builder.Default private final int ndots = DEFAULT_NDOTS;
+
+  /** Configures the search path used to look up relative names with less than ndots dots. */
+  @Singular("searchPathElement")
```
ibauersachs, line 65: This name confused me. Either make it searchPaths/searchPath or both searchPath

src/main/java/org/xbill/DNS/lookup/LookupSession.java
```diff
@@ -0,0 +1,299 @@
...
+  private final List<Name> searchPath;
+
+  /**
+   * If set to true, cached results with multiple records will be returned with the starting point
```
ibauersachs, line 69: `{@code true}`

src/main/java/org/xbill/DNS/lookup/LookupSession.java
```diff
@@ -0,0 +1,299 @@
...
+   */
+  @Builder.Default private final int maxRedirects = DEFAULT_MAX_ITERATIONS;
+  /**
+   * Sets the threshold for the number of dots which much appear in a name before it is considered
+   * absolute. The default is {@code 1}, meaning meaning that if there are any dots in a name, the
+   * name will be tried first as an absolute name.
+   */
+  @Builder.Default private final int ndots = DEFAULT_NDOTS;
+
+  /** Configures the search path used to look up relative names with less than ndots dots. */
```
ibauersachs, line 64: `{@link #ndots()}`

src/main/java/org/xbill/DNS/lookup/LookupSession.java
```diff
@@ -0,0 +1,299 @@
...
+  /**
+   * Make an asynchronous lookup of the provided name.
+   *
+   * @param name the name to look up.
+   * @param type the type to look up, values should correspond to constants in {@link Type}.
+   * @param dclass the class to look up, values should correspond to constants in {@link DClass}.
+   * @return A {@link CompletionStage} what will yield the eventual lookup result.
+   */
+  public CompletionStage<LookupResult> lookupAsync(Name name, int type, int dclass) {
+    CompletableFuture<LookupResult> future = new CompletableFuture<>();
```
ibauersachs, line 111: Create this future closer to its usage to avoid creation in case of failure

src/main/java/org/xbill/DNS/lookup/LookupSession.java
```diff
@@ -0,0 +1,299 @@
...
+                future.complete(result);
+              } else {
+                lookupUntilSuccess(names, type, dclass, future);
+              }
+            });
+  }
+
+  private CompletionStage<LookupResult> lookupWithCache(Record queryRecord, List<Name> aliases) {
+
+    Supplier<CompletionStage<LookupResult>> lookupWithResolver =
```
ibauersachs, line 179: I think it would be easier to read if this were a proper method

src/main/java/org/xbill/DNS/lookup/LookupSession.java
```diff
@@ -0,0 +1,299 @@
...
+
+  private Message maybeAddToCache(Message message) {
+    Optional.ofNullable(caches.get(message.getQuestion().getDClass()))
+        .ifPresent(cache -> cache.addMessage(message));
+    return message;
+  }
+
+  private CompletionStage<LookupResult> setResponseToMessageFuture(SetResponse setResponse) {
+    if (setResponse.isNXDOMAIN()) {
+      return completedFuture(new LookupResult(emptyList(), AdditionalDetail.NXDOMAIN, null));
```
ibauersachs, line 200: I'm unused to static imports except for assertions and (sometimes) constants

src/main/java/org/xbill/DNS/lookup/RedirectOverflowException.java
```diff
@@ -0,0 +1,8 @@
+// SPDX-License-Identifier: BSD-2-Clause
+package org.xbill.DNS.lookup;
+
+public class RedirectOverflowException extends LookupFailedException {
```
ibauersachs, line 4: class-level Javadoc

ibauersachs (changes requested):
More nitpicks

nresare:
> 
> I don't think [exceptions] is so bad. In this flow I defintely expect to have a valid, usable answer or an exception, not something in between:
> 

Having now experimented a bit, I can throw and catch 100k exceptions in about ~200ms on my laptop. If someone is hurt by the performance implications of that, I think they might have other problems in their design.

Let me implement an exception based approach and we can discuss further on the basis of that.

src/main/java/org/xbill/DNS/lookup/LookupSession.java
nresare: While this particular code has changed, for the general case I am reading your statement that you are unused to the practice in the context of a change request as that you would rather me not doing that. While I could definitely make a compromise and remove the static imports, I think that it improves readability. Please tell me more about what you think, why and how strong this opinion is :)

src/main/java/org/xbill/DNS/lookup/RedirectOverflowException.java
nresare: Added

src/main/java/org/xbill/DNS/lookup/AdditionalDetail.java
nresare: If we decide to go with exceptions instead we get rid of this whole enum

src/main/java/org/xbill/DNS/lookup/InvalidZoneDataException.java
nresare: Added.

src/main/java/org/xbill/DNS/lookup/LookupResult.java
nresare: Added some more detail on the semantics, and capitalised.

src/main/java/org/xbill/DNS/lookup/LookupSession.java
nresare: That doesn't seem to work. At least it shows up as red for me in my ide. I'll make a note to look at that when I get around to fixing the javadoc to include lombok generated code.

src/main/java/org/xbill/DNS/lookup/LookupSession.java
nresare: I believe this is the closest that the future can be created. The construct of having an outer future and then a function doing some asynchronous work one or more times by a recursive call requires the outer future to be defined in the outer function. Also note that It's not necessarily call stack recursion as the recursive call is made as a result of an asynchronous operation, possibly on some worker thread.

src/main/java/org/xbill/DNS/lookup/LookupSession.java
nresare: Agreed. This was me simply breaking out some part of a needlessly long chain into a variable, but a function is much better. Changed.

src/main/java/org/xbill/DNS/lookup/LookupResult.java
nresare: I feel fairly strongly that this should be public because it simplifies testing. If a user of the library want to unit test their code it would be natural to mock a LookupSession. If the test code can not create `LookupResult` instances without resorting to reflection magic (or mocking the actual value object) that adds a lot of complexity to the test setup.

This is something that Apache Kafka suffers from, where the admin api returns value objects without public constructors, and it has been pretty annoying.

ibauersachs: Well, okay, but then please keep in mind that it becomes public API that cannot be changed anymore.

src/main/java/org/xbill/DNS/lookup/LookupSession.java
ibauersachs: Yes, this comment was meant to be somewhere else... can't find it anymore.

src/main/java/org/xbill/DNS/lookup/LookupSession.java
ibauersachs: That's probably Lombok, `{@link #$default$ndots()}` exists (but makes no sense to link to).

src/main/java/org/xbill/DNS/lookup/LookupSession.java
ibauersachs: I think it's confusing (with local method names). I'd refer to some guidelines, but I don't have any at hand. If you're used to it from previous work, then I guess I could live with it. I've just never seen static imports for methods used outside of test code.

ibauersachs:
> I have pushed a version that uses two new exceptions instead of empty list and an enum to signal NXDOMAIN/NXRRSET. If I may say so myself, I think that this is a better approach.

Yes, it feels natural now.

> At the same time getting rid of the complexity of handling NameTooLongException for the edge case that we are now handling in Name. Unfortunately we now have some code that is not testable, so coverage is down from 100%, but I have a theory of how to address this with a later change.

If you want to make it an exercise, great, but there's really no need to exaggerate :-)

src/main/java/org/xbill/DNS/lookup/LookupSession.java
```diff
@@ -0,0 +1,295 @@
...
+      Name name = question.getName().fromDNAME((DNAMERecord) firstAnswer);
+      return Record.newRecord(name, question.getType(), question.getDClass());
+    } catch (NameTooLongException e) {
+      throw new InvalidZoneDataException(
+          "DNAME redirect would result in a name that would be too long");
+    }
+  }
+
+  /** Returns a LookupResult if this response was a non-exceptional empty result, else null. */
+  private static LookupResult buildResult(Message answer, List<Name> aliases) {
```
ibauersachs, line 277: Consider if you really need to throw the exceptions or if you could just return a failed future (the stacktrace in the exceptions is useless and it would probably half your exception-throwing-testbench-time).

And it could be useful to add the query's Name/Type to the exception; the caller should know it, but if the failure/exception is blindly used for logging it's helpful.

src/main/java/org/xbill/DNS/lookup/LookupSession.java
nresare: Thanks! Updated.

src/main/java/org/xbill/DNS/lookup/LookupSession.java
nresare: I think it makes sense to keep it private as long as we are not using it anywhere else.

src/main/java/org/xbill/DNS/lookup/LookupSession.java
nresare: For the person reading that comment, I think we should skip the indirection.

nresare: ooh, now I thing I'm getting  `@value` tags. Seems useful, but I can't get it to work for some reason.

src/main/java/org/xbill/DNS/lookup/LookupSession.java
nresare: I'm happy to change to have both be searchPath (yay for method overloading!). I don't think that plural searchPaths makes sense as other examples such as `$PATH` has multiple elements but still is a singular thing.

src/main/java/org/xbill/DNS/lookup/LookupSession.java
nresare: This is not yet implemented. I'm thinking that there should be a static method that would give you a _default_ LookupSession with sensible defaults set, but with ~100 comments to this PR, I would prefer to have that land a bit later.

src/main/java/org/xbill/DNS/lookup/LookupSession.java
nresare: Oh, that was a solution I did not find in my own search. It seemed to me like the it is only possible to control the access level of all the generated method, not specifically the generated builder. But I will use the anonymous subclass solution instead.

src/main/java/org/xbill/DNS/lookup/LookupSession.java
nresare: One could move away from having regular synchronous flow in these methods and instead allow for a mechanism that returned either a value or an exception, but I think that might be the wrong tradeoff as it diverges from a very well established pattern of execution with a simple return value for the successful case and an Exception for the exceptional case, simply to save a ms or two. 

I'm not so sure about adding other potentially useful information to the exception that is also available up the call stack. It makes things a bit simpler to not add it at the very least :)

nresare:
> If you want to make it an exercise, great, but there's really no need to exaggerate :-)

I am happy to not have coverage when there is a good reason not to, but otherwise I think it is useful to strive for 100, and I think it makes for better code and interfaces as well.

nresare:
This PR became a bit unwieldy, but I believe that I have mostly addressed the concerns you raised. Please have another look :)

I made an attempt to have the lombok generated public code be part of the javadoc but it was more difficult than I initially anticipated, so I would prefer to have this land as a separate standalone PR. In the last two years I have become very acquainted with gradle, and doing pretty much anything with maven these days feels like a detour, but that is a discussion for another time.

src/main/java/org/xbill/DNS/lookup/LookupSession.java
```diff
@@ -0,0 +1,314 @@
...
+  public static final int DEFAULT_NDOTS = 1;
+  /** The {@link Resolver} to use to look up records. */
+  @NonNull private final Resolver resolver;
+  /**
+   * Sets the maximum number of CNAME or DNAME redirects allowed before lookups with fail with
+   * {@link RedirectOverflowException}. Defaults to {@code 16}.
+   */
+  @Builder.Default private final int maxRedirects = DEFAULT_MAX_ITERATIONS;
+  /**
+   * Sets the threshold for the number of dots which much appear in a name before it is considered
```
ibauersachs, line 58: much -> must

src/main/java/org/xbill/DNS/lookup/LookupSession.java
```diff
@@ -0,0 +1,314 @@
...
+   * absolute. The default is {@value #DEFAULT_NDOTS}, meaning meaning that if there are any dots in
```
ibauersachs, line 59: `meaning meaning`

src/main/java/org/xbill/DNS/lookup/LookupSession.java
```diff
@@ -0,0 +1,314 @@
...
+            searchPath.stream()
+                .map(
+                    name -> {
+                      try {
+                        return Name.concatenate(name, Name.root);
+                      } catch (NameTooLongException e) {
+                        throw new IllegalArgumentException("Search path name too long");
+                      }
+                    })
+                .collect(Collectors.toCollection(ArrayList::new));
```
ibauersachs, line 103: The comment probably got lost, but what about `searchPath.stream()...().forEach(this.searchPath::add);` instead of ArrayList::new? It would avoid the creation of a never used list.

src/main/java/org/xbill/DNS/lookup/LookupSession.java
```diff
@@ -0,0 +1,314 @@
+// SPDX-License-Identifier: BSD-2-Clause
+package org.xbill.DNS.lookup;
+
+import static java.lang.String.format;
+import static java.util.Collections.singletonList;
+import static java.util.concurrent.CompletableFuture.*;
```
ibauersachs, line 6: Wildcard imports are not allowed in g-j-f. Seems I forgot to bind `checkstyle:check` to the verify phase (and I'd still prefer not using static method imports at all).

src/main/java/org/xbill/DNS/lookup/LookupSession.java
```diff
@@ -0,0 +1,314 @@
...
+      copy.add(name);
+      return Collections.unmodifiableList(copy);
+    }
+  }
+
+  private Record buildRedirectQuery(LookupResult response, Record question) {
+    List<Record> answer = response.getRecords();
+    Record firstAnswer = answer.get(0);
+    if (answer.size() != 1) {
+      throw new InvalidZoneDataException("Multiple CNAME RRs not allowed, SEE RFC1034 3.6.2");
```
ibauersachs, line 278: `see` or is there a reason for writing this uppercase?

src/main/java/org/xbill/DNS/lookup/LookupSession.java
ibauersachs: setResponseToMessageFuture(...) does such a non-throwing, failed-future flow. But I agree that this is currently more readable.

There were a few issues that requested adding such information to exceptions.

ibauersachs (changes requested):
Some small stuff, and this as tbd for follow ups:
- Delombok for Javadoc in of the builder methods
- ndots from/interaction with ResolverConfig
- max iterations sytem property
- simplified lookupAsync(...) overload(s)
- simplified and/or default Cache
- create() with sensible defaults

src/main/java/org/xbill/DNS/lookup/LookupSession.java
```diff
@@ -0,0 +1,314 @@
...
+@Builder
+public class LookupSession {
+  // remember to update javadoc if this is changed for the following two properties.
+  public static final int DEFAULT_MAX_ITERATIONS = 16;
+  public static final int DEFAULT_NDOTS = 1;
+  /** The {@link Resolver} to use to look up records. */
+  @NonNull private final Resolver resolver;
+  /**
+   * Sets the maximum number of CNAME or DNAME redirects allowed before lookups with fail with
+   * {@link RedirectOverflowException}. Defaults to {@code 16}.
```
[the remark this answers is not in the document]
nresare: I got the `{@value VALUE}` construct to work. (It turns out that with the latest released java 8 javadoc tool, one need to have the fully qualified field name of the constant field, presumably since the javadoc field is copied to an inner class by lombok).

I agree that it makes sense to add reading of the default to the code that constructs the sensible default instance, along with setting up caches and so on.

src/main/java/org/xbill/DNS/lookup/LookupSession.java
```diff
@@ -0,0 +1,314 @@
...
+import org.xbill.DNS.Type;
+
+/**
+ * LookupSession provides facilities to make DNS Queries. A LookupSession is intended to be long
+ * lived, and it's behaviour can be configured using the properties of the LookupSessionBuilder
+ * instance returned by the builder() method.
+ */
+@Builder
+public class LookupSession {
+  // remember to update javadoc if this is changed for the following two properties.
```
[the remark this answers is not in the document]
nresare: Fixed.

src/main/java/org/xbill/DNS/lookup/LookupSession.java
nresare: fixed.

src/main/java/org/xbill/DNS/lookup/LookupSession.java
nresare: fixed.

src/main/java/org/xbill/DNS/lookup/LookupSession.java
nresare: Moved to spelling out the classes instead

src/main/java/org/xbill/DNS/lookup/LookupSession.java
nresare: Well, calling add() on a list from within a stream over that list causes a `ConcurerntModificationException`. While it is good to not create throwaway values left and right, I think that constructing a new list is okay in this case. Let's call it the least worst option.

src/main/java/org/xbill/DNS/lookup/LookupSession.java
nresare: No reason 🤷

src/main/java/org/xbill/DNS/lookup/LookupSession.java
ibauersachs: Ah right, I didn't pay attention that this is the same list.

ibauersachs (approved):

nresare:
Thank you Ingo for putting so much effort into this review. I will try to make the upcoming changes smaller and easier to review :)

PR #164, opened 2021-02-14
Some cleanups

ibauersachs:


nresare (approved):
Sure, looks good to me. I am assuming that here is some good reason for dropping android < 26 support

ibauersachs:
[java.time.Instant](https://developer.android.com/reference/java/time/Instant) / java.time.Duration were only added in API level 26, similarly, [java.util.concurrent.CompletionStage](https://developer.android.com/reference/java/util/concurrent/CompletionStage) is only there since level 24.

PR #165, opened 2021-02-15
Have constructing relative Name with length() == MAXNAME fail

nresare:
The Name() constructor will currently fail if it's length() would
end up being longer than MAXNAME (255). However, there is an edge
case where a relative name can be exactly MAXNAME long. This is
legal but the lookup mechanism will then fail to make the name
absolute by appending an empty label to be able to use it for
lookups.

### DIFF
--- a/src/test/java/org/xbill/DNS/NameTest.java
+++ b/src/test/java/org/xbill/DNS/NameTest.java
@@ -34,6 +34,7 @@
 //
 package org.xbill.DNS;
 
+import static java.lang.String.format;
 import static org.junit.jupiter.api.Assertions.assertArrayEquals;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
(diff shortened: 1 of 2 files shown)


src/test/java/org/xbill/DNS/NameTest.java
```diff
@@ -286,15 +289,26 @@ void ctor_toobig_label() {
...
-    void ctor_max_length_rel() throws TextParseException {
-      // relative name with three 63-char labels and a 62-char label
+    void ctor_length_too_long_rel() {
+      // We want to fail to crate this as there is no way to make a Name that long absolute,
+      // which means that it can not be used for lookups or updates.
+      String label63 = IntStream.range(0, 63).mapToObj(i -> "a").collect(Collectors.joining());
+      String label62 = IntStream.range(0, 62).mapToObj(i -> "a").collect(Collectors.joining());
+      assertThrows(
+          TextParseException.class,
+          () -> new Name(format("%s.%s.%s.%s", label63, label63, label63, label62)));
```
ibauersachs, line 299: I'm not used to importing anything statically other than assertions or (sometimes) constants.

ibauersachs (approved):
I hope this doesn't break stuff. But comparing it with [ldns](https://github.com/NLnetLabs/ldns), I think it's okay. Although ldns allows creating a 3x63+62 relative name with status OK, it fails to do anything with it later on, including to converting back to string. ldns even violates it's own guarantee that an rdf internally is always absolute in the 3x63+62 case and returns a length of 256.

PR #168, opened 2021-02-20
Fix unhandled NPE

nekohasekai:


nresare (approved):
This seems like a sensible change. The [javadoc for HttpURLConnection ](https://docs.oracle.com/javase/8/docs/api/java/net/HttpURLConnection.html#getErrorStream--) clearly state that `.getErrorStream()` might be null.

src/main/java/org/xbill/DNS/DohResolver.java
```diff
@@ -366,10 +366,13 @@ public Duration getTimeout() {
         }
       }
     } catch (IOException ioe) {
-      try (InputStream es = conn.getErrorStream()) {
-        byte[] buf = new byte[4096];
-        while (es.read(buf) > 0) {
```
ibauersachs, line 371: It'd be enough/easier to read to write this as ` while (es != null && es.read(buf) > 0) {`. This works because try-with-resources handles `null` gracefully.

PR #170, opened 2021-02-23
Introduce getters for LookupFailedExcpetion properties name and type

nresare:
Oops. Forgot this.

### DIFF
--- a/src/test/java/org/xbill/DNS/lookup/LookupFailedExceptionTest.java
+++ b/src/test/java/org/xbill/DNS/lookup/LookupFailedExceptionTest.java
@@ -1,0 +1,39 @@
+// SPDX-License-Identifier: BSD-2-Clause
+package org.xbill.DNS.lookup;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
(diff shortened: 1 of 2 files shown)


src/test/java/org/xbill/DNS/lookup/LookupFailedExceptionTest.java
```diff
@@ -0,0 +1,38 @@
+package org.xbill.DNS.lookup;
```
ibauersachs, line 1: Please add the SPDX license header

ibauersachs (approved):
I missed that the header is missing on some of the new files of #162 as well:
- LookupSessionTest.java
- NoSuchRRSetException.java
- NoSuchDomainException.java 

We should probably add the header to all existing files as well, regardless of what is already there and enforce it using a checkstyle rule.

src/test/java/org/xbill/DNS/lookup/LookupFailedExceptionTest.java
nresare: Sure. Added.

nresare:
That sounds like a plan :)

PR #171, opened 2021-02-24
Enforce license info header

nresare:
I split the check for the licensing header and the adding of the actual headers to the files into two commits to make it as little less unwieldy.


ibauersachs (approved):
Not sure I like the config being external to the pom, but it also doesn't bother me enough to really care.

Thanks for the legwork!

nresare:
Having xml config inside the POM and xml config outside the POM is two kinds of bad to me, with a weak preference for splitting out as much as possible from the pom. But it is good I know your preference now, and will take it into account next time :)

PR #173, opened 2021-02-25
Expand wildcard names in Zone lookups

ibauersachs:


src/main/java/org/xbill/DNS/Zone.java
```diff
@@ -337,25 +337,18 @@ private synchronized void removeRRset(Name name, int type) {
   }
 
   private synchronized SetResponse lookup(Name name, int type) {
-    int labels;
```
nresare, line 340: It is almost as if the original author comes from a C background :)

src/test/java/org/xbill/DNS/ZoneTest.java
```diff
@@ -0,0 +1,117 @@
...
+
+import java.io.IOException;
+import java.net.InetAddress;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+class ZoneTest {
+  private Zone zone;
+
+  @BeforeEach
```
nresare, line 15: Since this data is read-only, `@BeforeAll` would work equally well

src/test/java/org/xbill/DNS/ZoneTest.java
```diff
@@ -0,0 +1,117 @@
...
+          new TXTRecord(new Name("*", nameZone), DClass.IN, 3600, "sometext")
+        };
+    zone = new Zone(nameZone, zoneRecords);
+  }
+
+  @Test
+  void exactNameExistingALookup() {
+    Name testName = Name.fromConstantString("test.example.");
+    SetResponse resp = zone.findRecords(testName, Type.A);
+    assertEquals(testName, resp.answers().get(0).first().getName());
```
nresare, line 48: In general I think that leaning on the the excellent list comparison abilities of `assertEquals` is preferable to looking at specific properties of the returned answer, but since `RRset` doesn't have a succinct way of creating an instance and is also missing equals and hash methods there would be some work to get that into the code.

In this case though, the tests are such a clear improvement so feel free to land.

nresare (approved):

src/test/java/org/xbill/DNS/ZoneTest.java
ibauersachs: equals/hashCode of rrset is a one-liner, so I tried this now. It makes constructing the zone much more awkward with the intermediate construction of RRsets and the only test that really benefits is exactNameAnyLookup. I've created  #174 for comparison. Have a look please, I'm not sure what I like better.

nresare: Oh interesting. I'll comment over there.

ibauersachs:
Superseded by #174

PR #174, opened 2021-02-27
Expand wildcard names in Zone lookups

### LINKED ISSUE #169: Zone.lookup fails or behaves badly when using wildcards
Hi, i'm using dnsjava in a small authoritative server project (https://github.com/thomaskuh/limbodns) and got 2 issues resolving requests when zone has a wildcard record:

- When lookin up a name handled by a wildcard, the result isn't substituted (Returns *.abc. instead or requestedname.abc.).
- When lookin up a name with type ANY an exception is thrown.

Here's a little test I wrote showing the (in my option) wrong behaviour:
```
public static final void main(String[] args) throws Exception {
	Name nameZone = new Name("example.");
	Zone zone = new Zone(nameZone, new Record[] {
		new SOARecord(nameZone, DClass.IN, 3600L, Name.fromString("nameserver."), new Name("hostmaster",nameZone), 1, 21600L, 7200L, 2160000L, 3600L),
		new NSRecord(nameZone, DClass.IN, 300L, Name.fromString("nameserver.")),
		new ARecord(new Name("test", nameZone), DClass.IN, 3600, Inet4Address.getLocalHost()),
		new ARecord(new Name("*", nameZone), DClass.IN, 3600, Inet4Address.getLocalHost()),
		new TXTRecord(new Name("*", nameZone), DClass.IN, 3600, "sometext")
	});
	
	SetResponse resp = null;
	resp = zone.findRecords(Name.fromString("test.example."), Type.A);
	System.out.println(resp + " -> " + resp.answers());
	// successful -> [{ test.example. 3600 IN A [10.0.0.214] }]
	
	resp = zone.findRecords(Name.fromString("undefined.example."), Type.A);
	System.out.println(resp + " -> " + resp.answers());
	// successful -> [{ *.example. 3600 IN A [10.0.0.214] }]
	// Seems to work but i'd expect:
	// successful -> [{ undefined.example. 3600 IN A [10.0.0.214] }]
	
	resp = zone.findRecords(Name.fromString("undefined.example."), Type.TXT);
	System.out.println(resp + " -> " + resp.answers());
	// successful -> [{ *.example. 3600 IN TXT ["sometext"] }]
	// Seems to work but i'd expect:
	// successful -> [{ undefined.example. 3600 IN TXT ["sometext"] }]
	
	resp = zone.findRecords(Name.fromString("undefined.example."), Type.ANY);
	System.out.println(resp + " -> " + resp.answers());
	// FAILS WITH: java.lang.IllegalArgumentException: oneRRset(ANY) at org.xbill.DNS.Zone.oneRRset(Zone.java:249)
	// Would expect:
	// successful -> [{ *.example. 3600 IN A [10.0.0.214] }, { *.example. 3600 IN TXT ["sometext"] }]
	// or even better:
	// successful -> [{ undefined.example. 3600 IN A [10.0.0.214] }, { undefined.example. 3600 IN TXT ["sometext"] }]
}
```

ibauersachs:
Closes #169

### DIFF
--- a/src/main/java/org/xbill/DNS/RRset.java
+++ b/src/main/java/org/xbill/DNS/RRset.java
@@ -8,6 +8,8 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Iterator;
 import java.util.List;
+import java.util.Objects;
+import lombok.EqualsAndHashCode;
 
 /**
  * A set of Records with the same name, type, and class. Also included are all RRSIG records signing
@@ -17,6 +19,7 @@ import java.util.List;
  * @see RRSIGRecord
  * @author Brian Wellington
  */
+@EqualsAndHashCode(of = {"rrs", "sigs"})
 public class RRset implements Serializable {
   private final ArrayList<Record> rrs;
   private final ArrayList<RRSIGRecord> sigs;
@@ -35,6 +38,19 @@ public class RRset implements Serializable {
     addRR(record);
   }
 
+  /**
+   * Creates an RRset and sets its contents to the specified record(s)
+   *
+   * @param records The records to add to the set. See {@link #addRR(Record)} for restrictions.
+   */
+  public RRset(Record... records) {
+    this();
+    Objects.requireNonNull(records);
+    for (Record r : records) {
+      addRR(r);
+    }
+  }
+
   /** Creates an RRset with the contents of an existing RRset */
   public RRset(RRset rrset) {
     rrs = new ArrayList<>(rrset.rrs);
@@ -47,6 +63,9 @@ public class RRset implements Serializable {
    * Adds a signature to this RRset. If the TTL of the added signature is not the same as existing
    * records in the RRset, all records are set to the lowest TTL of either the added record or the
    * existing records.
+   *
+   * @throws IllegalArgumentException if the RRset already contains records and the signature to add
+   *     does not match.
    */
   public void addRR(RRSIGRecord r) {
     addRR(r, sigs);
@@ -56,6 +75,9 @@ public class RRset implements Serializable {
    * Adds a Record to this RRset. If the TTL of the added record is not the same as existing records
    * in the RRset, all records are set to the lowest TTL of either the added record or the existing
    * records.
+   *
+   * @throws IllegalArgumentException if the RRset already contains records and the record to add
+   *     does not match.
    */
   public void addRR(Record r) {
     if (r instanceof RRSIGRecord) {
--- a/src/test/java/org/xbill/DNS/ZoneTest.java
+++ b/src/test/java/org/xbill/DNS/ZoneTest.java
@@ -1,0 +1,120 @@
+// SPDX-License-Identifier: BSD-2-Clause
+package org.xbill.DNS;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import java.io.IOException;
+import java.net.InetAddress;
+import java.util.Collections;
+import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+import org.junit.jupiter.api.Test;
+
+class ZoneTest {
+  private static final ARecord A_TEST;
+  private static final AAAARecord AAAA_1_TEST;
+  private static final AAAARecord AAAA_2_TEST;
+  private static final ARecord A_WILD;
+  private static final TXTRecord TXT_WILD;
+  private static final Zone ZONE;
+
+  static {
+    try {
+      Name nameZone = new Name("example.");
+      InetAddress localhost4 = InetAddress.getByName("127.0.0.1");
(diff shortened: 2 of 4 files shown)


nresare:
While this certainly works, I would prefer to move towards treating `RRset` as one would an immutable data structure, and instead of using `addRR()` have either a constructor that takes the records or perhaps a more localised solution with a static method that constructs a `RRset` from one or more Record.

I realise we are heavily into opinion and taste territory here, and I am happy to go with whatever solution you have, but since you are asking, I'm happy to share my mine. Feel free to disregard.

I would have done the following:

Created a method to create rrset instances more easily, something along the lines of

```java
private static RRset rrset(Record... records) {
    RRset set = new RRset();
    Arrays.stream(records).forEach(set::addRR);
    return set;
}
```
(for future consideration including into `RRest` as a constructor or static factor method)

Armed with this I would create the `RRset` lists that goes into the first arg of `assertEquals()` inline when I needed them. Possibly, just having the different `Record` instances as `private static final` class fields.

nresare:
My bias is to define the test data that goes into a unit test inline in the actual test most of the time, potentially using an `private static final Type UPPERCASE` constant declaration for values. 

But I do appreciate having something like `assertEquals(singletonList(rrset(LOCALHOST4)), resp.answers())` instead of having the full chain like `assertEquals(1, resp.answers().get(0).rrs().size())` for these reasons:

* The error messages when things are wrong become more descriptive, especially when `toString()` is well written.
* If one want's to be thorough, there are quite a few assert lines that makes sure that the `answers()` method returns exactly the right thing.n
* Sometimes there might be a zero length list return value and the assertEquals might assume that `get(0)` will work and the test failure then becomes an `IndexOutOfBoundsException` which is almost always less helpful than a diff of the `toString()` of the expected and actual values going into the assertion

ibauersachs:
> While this certainly works, I would prefer to move towards treating `RRset` as one would an immutable data structure, and instead of using `addRR()` have either a constructor that takes the records or perhaps a more localised solution with a static method that constructs a `RRset` from one or more Record.

I'd design it like that today, but we can't remove the mutable API. An additional constructor for multiple records is a good idea.

> I realise we are heavily into opinion and taste territory here, and I am happy to go with whatever solution you have, but since you are asking, I'm happy to share my mine. Feel free to disregard.
> 
> I would have done the following:
> 
> Created a method to create rrset instances more easily, something along the lines of
> 
> ```java
> private static RRset rrset(Record... records) {
>     RRset set = new RRset();
>     Arrays.stream(records).forEach(set::addRR);
>     return set;
> }
> ```
> 
> (for future consideration including into `RRest` as a constructor or static factor method)

I prefer the constructor over the static method and added it.

> Armed with this I would create the `RRset` lists that goes into the first arg of `assertEquals()` inline when I needed them. Possibly, just having the different `Record` instances as `private static final` class fields.
>
> My bias is to define the test data that goes into a unit test inline in the actual test most of the time, potentially using an `private static final Type UPPERCASE` constant declaration for values.

The checked exception unfortunately demand a static constructor for that. Almost the same as beforeAll, but at least the fields can be final.

> But I do appreciate having something like `assertEquals(singletonList(rrset(LOCALHOST4)), resp.answers())` instead of having the full chain like `assertEquals(1, resp.answers().get(0).rrs().size())` for these reasons:
> 
> * The error messages when things are wrong become more descriptive, especially when `toString()` is well written.
> * If one want's to be thorough, there are quite a few assert lines that makes sure that the `answers()` method returns exactly the right thing.n
> * Sometimes there might be a zero length list return value and the assertEquals might assume that `get(0)` will work and the test failure then becomes an `IndexOutOfBoundsException` which is almost always less helpful than a diff of the `toString()` of the expected and actual values going into the assertion

Yes, it's perfectly clear why.

src/test/java/org/xbill/DNS/ZoneTest.java
```diff
@@ -0,0 +1,120 @@
...
+    assertEquals(
+        listOf(new RRset(A_WILD.withName(testName)), new RRset(TXT_WILD.withName(testName))),
+        resp.answers());
+  }
+
+  private static List<RRset> listOf(RRset... rrsets) {
+    return Stream.of(rrsets).collect(Collectors.toList());
+  }
+
+  private static List<RRset> oneRRset(Record... r) {
```
nresare, line 117: 👍

src/main/java/org/xbill/DNS/RRset.java
```diff
@@ -35,6 +37,16 @@ public RRset(Record record) {
     addRR(record);
   }
 
+  /** Creates an RRset and sets its contents to the specified record(s) */
+  public RRset(Record... records) {
+    this();
+    if (records != null) {
```
nresare, line 43: The java vararg design is terrible in that it is possible to pass in a null value of type Record[]. While I think this is fine, I like the `Objects.requireNoNull()` construct instead. If someone explicitly feeds this method a `(Record[])null` they deserve a NullPointerException ;)

nresare (approved):

PR #179, opened 2021-03-22
Fix load balancing

paulo-raca:
`lbStart.updateAndGet(i -> i++ % resolvers.size())` is a no-op:

`i++` increments the local variable, which is lost, and the returned value is simply `i % resolvers.size()`


ibauersachs:
Good catch, but the fix is wrong: modulo has higher precedence than addition. Your version is thus effectively `1 % resolvers.size()`. Correct would be `++i % resolvers.size()` or `(i + 1) % resolvers.size()`. Do you want to update this PR?

paulo-raca:
🤦 
Fixed now, thanks!

Issue #182, opened 2021-03-27
Java Native Support


YunaBraska:
Hi I am currently testing java native with Quarkus using GraalVM and get the following stack trace.
It might be the issue with the `SecureRandom`. I have not much knowledge about `SecureRandom` is it really that different than the usual `Random`? Else can you have a look into this? I have no clue how to fix this but would really like to use the library :D Also Spring Boot goes native as well it might be a good time to make this library work on java native :) 

```
Error: No instances of java.net.Inet4Address are allowed in the image heap as this class should be initialized at image runtime. To see how this object got instantiated use --trace-object-instantiation=java.net.Inet4Address.
Trace: Object was reached by 
        reading field java.net.InetSocketAddress$InetSocketAddressHolder.addr of
                constant java.net.InetSocketAddress$InetSocketAddressHolder@4e2cb91a reached by 
        reading field java.net.InetSocketAddress.holder of
                constant java.net.InetSocketAddress@5d76dcb9 reached by 
        reading field org.xbill.DNS.SimpleResolver.address of
                constant org.xbill.DNS.SimpleResolver@ba41067 reached by 
        reading field org.xbill.DNS.ExtendedResolver$ResolverEntry.resolver of
                constant org.xbill.DNS.ExtendedResolver$ResolverEntry@3cefc7af reached by 
        indexing into array
                constant java.lang.Object[]@2c7498c0 reached by 
        reading field java.util.concurrent.CopyOnWriteArrayList.array of
                constant java.util.concurrent.CopyOnWriteArrayList@3105f154 reached by 
        reading field org.xbill.DNS.ExtendedResolver.resolvers of
                constant org.xbill.DNS.ExtendedResolver@27f6e74 reached by 
        reading field org.xbill.DNS.Lookup.defaultResolver
Error: No instances of sun.security.provider.NativePRNG are allowed in the image heap as this class should be initialized at image runtime. To see how this object got instantiated use --trace-object-instantiation=sun.security.provider.NativePRNG.
Trace: Object was reached by 
        reading field java.security.SecureRandom.secureRandomSpi of
                constant java.security.SecureRandom@2048aa39 reached by 
        scanning method org.xbill.DNS.Header.getID(Header.java:140)
Call path from entry point to org.xbill.DNS.Header.getID(): 
        at org.xbill.DNS.Header.getID(Header.java:140)
        at org.xbill.DNS.Header.toStringWithRcode(Header.java:253)
[...]

PR #188, opened 2021-04-13
Add support for EDNS Extended Error Codes, RFC 8914

ibauersachs:


gryphius:
Thanks, this is working as expected. I'm testing this with the latest PowerDNS Recursor Beta.
Relevant resolver config:

recursor.conf:
```
local-address=192.168.23.100:5300
dnssec=validate
lua-config-file=/etc/pdns-recursor/config.lua
extended-resolution-errors=yes
```
config lua:
```
rpzFile("/etc/pdns-recursor/localoverride.rpz", {extendedErrorCode=15, extendedErrorExtra="blocked by local test RPZ" } )
```
localoverride.rpz:
```
$TTL 1h;
$ORIGIN domain.example.com.
@          SOA powerdns.example.net. hostmaster.example.com ( 1 12h 15m 3w 2h)
           NS powerdns.example.net.
; begin RPZ RR definitions
dnsjavatest.example.com CNAME .
```

test program:
```java
import org.xbill.DNS.*;

public class EDETest{
    public static void main(String[] args)throws Exception{
       String qname="dnsjavatest.example.com.";
       org.xbill.DNS.Record question_record = org.xbill.DNS.Record.newRecord(org.xbill.DNS.Name.fromString(qname), 1, DClass.IN);
       Message query = Message.newQuery(question_record);
        SimpleResolver resolver = new SimpleResolver("192.168.23.100");
        resolver.setPort(5300);
        Message response = resolver.send(query);
        System.out.println(response);
    }
}
```

the result shows the correct textual description and the extra string, as expected:

```
;; ->>HEADER<<- opcode: QUERY, status: NXDOMAIN, id: 15670
;; flags: qr rd ra ; qd: 1 an: 0 au: 0 ad: 1 
;; QUESTIONS:
;;	dnsjavatest.example.com., type = A, class = IN

;; ANSWERS:

;; AUTHORITY RECORDS:

;; ADDITIONAL RECORDS:
.			0	CLASS512	OPT	[{Extended_DNS_Error: Blocked: blocked by local test RPZ}]  ; payload 512, xrcode 0, version 0, flags 0

;; Message size: 83 bytes

```

src/main/java/org/xbill/DNS/ExtendedErrorCodeOption.java
```diff
@@ -0,0 +1,133 @@
...
+  @Override
+  void optionToWire(DNSOutput out) {
+    out.writeU16(errorCode);
+    if (text != null && text.length() > 0) {
+      out.writeByteArray(text.getBytes(StandardCharsets.UTF_8));
+    }
+  }
+
+  @Override
+  String optionToString() {
```
hyperxpro, line 127: Can we simply call this method `toString()`?

ibauersachs: No, it overrides a method (which is already public API). The base class already has `toString()` overridden:
https://github.com/dnsjava/dnsjava/blob/892fab2d826e6561c7c28b69461f3a965f2b397c/src/main/java/org/xbill/DNS/EDNSOption.java#L119-L122

Which e.g. produces `{Extended_DNS_Error: DNSSEC Indeterminate: not signed}`. I could still change the formatting of `optionToString()` and e.g. `Extended_DNS_Error` to `EDE` if that would be preferred.

nresare (approved):
This looks good to me. 

In general I am no fan of static initialisation blocks, but getting away from that for the whole code base is a bit of a task 🙃 If there was not a precedent already set in the use of static blocks, I would create a static accessor method that lazily initialises the structure when called.

Also, should we create create a helper method once and for allt o generate byte[] from a variable number of integers, handling the casting and signedness issues with byte?

ibauersachs:
@nresare I guess we could refactor the static stuff at some point. Unfortunately it's not that easy to do lazy thread safe init in Java, lacking something like C#'s `Lazy<T>`. OTOH, most static blocks are just a map init, and those would actually be beneficial to AOT compilation for GraalVM, as asked in #182. But we should definitely get rid of anything non-deterministic.

Yeah, why not. I guess this can live in the unit test for the time being, there isn't much casting or construction of byte[] in the main code as that comes from the network. I merged anyway, this can go into a dedicated refactoring PR.

Issue #186, opened 2021-04-13
HOST_NOT_FOUND is being cached indefinitely


amitknx:
Hi,
I am using dnsjava 3.1 to check the name server records for a domain and found out that
if the domain does not exist it is throwing HOST_NOT_FOUND but even after creating ns records it is still throwing HOST_NOT_FOUND.
What is the default caching ttl in case of HOST_NOT_FOUND?
when I restarted the service it got resolved?

ibauersachs:
First of all, please use the newest version, which is currently 3.3.1. And I cannot quite follow your question. Please attach the full nameserver reply that you think is incorrectly handled, with example code of how you're using dnsjava, the actual outcome and the expected outcome. A good example for such a report is #169.

amitknx:
Sorry for the incomplete explanation
Following is the code I am using:
`public boolean isZoneSetup(Name subDomain) {
        log.info(IS_ZONE_SETUP, subDomain);
        try {
           List<String> nameservers = List.of("8.8.8.8", "8.8.4.4");
            ExtendedResolver resolver = new ExtendedResolver(nameServers.toArray(String[]::new));
            resolver.setRetries(3);
            Lookup domainLookup = new Lookup(subDomain, Type.NS);
            domainLookup.setResolver(resolver);
            Record[] nsRecords = domainLookup.run();
            int lookupResult = domainLookup.getResult();
            switch (lookupResult) {
                case Lookup.SUCCESSFUL:
                    return true;
                case Lookup.TRY_AGAIN:
                    return false;
                case Lookup.HOST_NOT_FOUND:
                    throw new DomainNotFoundInDNSException(context);
                case Lookup.TYPE_NOT_FOUND:
                    throw new DNSRecordNotFoundException(context);
                case Lookup.UNRECOVERABLE:
                    throw new DNSLookupFailedException(context);
                default:
                    if (domainLookup.getErrorString().equals("SERVFAIL")) return false;
                    throw new DNSLookupFailedException(context);
            }
        } catch (UnknownHostException uhe) {
            log.error(uhe)
           throw uhe;
        }
    }`

- Calling first time for subdomain (abc.amitknx.com) with no nameserver records setup:
           Expected: HOST_NOT_FOUND
           Actual: HOST_NOT_FOUND

- Setup nameservers for abc.amitknx.com:
           Expected: SUCCESSFUL
           Actual: HOST_NOT_FOUND even after an hour

- After Restarting service: result SUCCESSFUL

[...]

PR #191, opened 2021-04-18
Fixed ttl for soa reacord to minimum of ttl and minimum field

amitknx:
I have been seeing an error where in case of negative answer(NXDOMAIN), ttl of cache is being set at MINIMUM field of the SOA record (see [issue#186](https://github.com/dnsjava/dnsjava/issues/186)) whereas [RFC2308](https://tools.ietf.org/html/rfc2308), Section 3 says:

> The TTL of this record is set from the minimum of the MINIMUM field of the SOA record and the TTL of the SOA itself, and indicates how long a resolver may cache the negative answer.

I have explore the code and found out that it indeed is being set for MINIMUM field, I have added a fix for it to set the cache ttl at minimum of the MINIMUM field of the SOA record and the TTL of the SOA.

### DIFF
--- a/src/main/java/org/xbill/DNS/Cache.java
+++ b/src/main/java/org/xbill/DNS/Cache.java
@@ -398,7 +398,7 @@ public class Cache {
   public synchronized void addNegative(Name name, int type, SOARecord soa, int cred) {
     long ttl = 0;
     if (soa != null) {
-      ttl = soa.getTTL();
+      ttl = Math.min(soa.getMinimum(), soa.getTTL());
     }
     Element element = findElement(name, type, 0);
     if (ttl == 0) {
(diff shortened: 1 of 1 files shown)


ibauersachs:
Yes, you're right. I only looked at `org.xbill.DNS.Cache#addNegative`, which needs the same `Math.min()` call. Would you mind updating the PR so that the minimum is considered in both checks?

ibauersachs (changes requested):
Yes, you're right. I only looked at org.`xbill.DNS.Cache#addNegative`, which needs the same `Math.min()` call. Would you mind updating the PR so that the minimum is considered in both checks?

amitknx:
Sorry, my mistake.. updated the PR

ibauersachs:
Thanks!

amitknx:
@ibauersachs : Hi, When can I expect this to be released in 3.4? Is there a timeline?

ibauersachs:
@amitknx I was hoping to get reviews for two open PRs from @nresare. I'll wait another week or two and would release anyway then.

amitknx:
ok thanks for the info.

PR #192, opened 2021-04-18
Add a method to shutdown the network I/O

### LINKED ISSUE #180: Clean shutdown when run inside Tomcat
I am using dnsjava inside a web-app running under Tomcat and I am getting the following error when the web-app is reloaded.

`24-Mar-2021 11:03:16.941 WARNING [Catalina-utility-2] org.apache.catalina.loader.WebappClassLoaderBase.clearReferencesThreads The web application [ROOT] appears to have started a thread named [dnsjava NIO selector] but has failed to stop it. This is very likely to create a memory leak. Stack trace of thread:
 java.base@11.0.9.1/sun.nio.ch.KQueue.poll(Native Method)
 java.base@11.0.9.1/sun.nio.ch.KQueueSelectorImpl.doSelect(KQueueSelectorImpl.java:122)
 java.base@11.0.9.1/sun.nio.ch.SelectorImpl.lockAndDoSelect(SelectorImpl.java:124)
 java.base@11.0.9.1/sun.nio.ch.SelectorImpl.select(SelectorImpl.java:136)
 org.xbill.DNS.Client.runSelector(Client.java:67)
 org.xbill.DNS.Client$$Lambda$291/0x0000000800493c40.run(Unknown Source)
 java.base@11.0.9.1/java.lang.Thread.run(Thread.java:834)`

It appears as though the Client selector and associated thread is closed in the shutdown hook. Is there a way to manually close this during the web-app unloading?

I've never written a PR before but I'm happy to try adding a manual close() if necessary.

ibauersachs:
Closes #180

### DIFF
--- a/src/main/java/org/xbill/DNS/Lookup.java
+++ b/src/main/java/org/xbill/DNS/Lookup.java
@@ -215,7 +215,7 @@ public final class Lookup {
    * @param logger The logger
    */
   public static synchronized void setPacketLogger(PacketLogger logger) {
-    Client.setPacketLogger(logger);
+    NioClient.setPacketLogger(logger);
   }
 
   private void reset() {
--- a/src/main/java/org/xbill/DNS/NioClient.java
+++ b/src/main/java/org/xbill/DNS/NioClient.java
@@ -11,19 +11,31 @@ import java.nio.channels.Selector;
 import java.util.Iterator;
 import java.util.List;
 import java.util.concurrent.CopyOnWriteArrayList;
+import lombok.AccessLevel;
+import lombok.NoArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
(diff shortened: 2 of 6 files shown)


ibauersachs:
@nresare I thought about refactoring this more, following your input in #180: Keep NioClient, but make everything in there non-static, implement AutoCloseable and have a lazy accessor for existing constructors of SimpleResolver. Add a new constructor in SimpleResolver accepting an instance of NioClient.
But that would require merging the TPC/UDP class into the base because they share the selector thread, and I'm not convinced it's a good idea of having multiple NioClients (and thus selectors and threads) around.

Until we have a better idea to address the resource closing, this is IMO the best solution to address #180.

nresare (approved):
Looks good to me!

PR #200, opened 2021-06-16
fixes to allow signing with ED25519 and ED448 algorithms

demonti:
Hi Ingo,
I discovered that dnsjava was unable to create signatures for the ED25519 and ED448 algorithms. This pull request contains the fixes to make them work.

### DIFF
--- a/src/main/java/org/xbill/DNS/DNSSEC.java
+++ b/src/main/java/org/xbill/DNS/DNSSEC.java
@@ -1119,12 +1119,12 @@ public class DNSSEC {
         }
         break;
       case Algorithm.ED25519:
-        if (!"Ed25519".equals(key.getAlgorithm())) {
+        if (!"Ed25519".equals(key.getAlgorithm()) && !"EdDSA".equals(key.getAlgorithm()) {
           throw new IncompatibleKeyException();
         }
         break;
       case Algorithm.ED448:
-        if (!"Ed448".equals(key.getAlgorithm())) {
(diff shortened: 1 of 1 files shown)


ibauersachs:
Thanks! Would you mind creating a unit test that covers ED signing?

And I'm a bit confused that `EdDSA` was working for verification. Are the JCA providers just intelligent enough to do a `verify` with EdDSA as algorithm name, regardless of whether Ed448 or Ed25519 is actually used?

demonti:
I have not discovered any tests for any other algorithm, so I did not create one. If I missed them, please point me to them! I have tested the changes with our name server product against the LDNS drill tool, and it could verify the generated signatures.

In respect to EdDSA vs Ed25519 and Ed448: I guess that they distinguish the algorithms by their key length. I have taken a look into RFC8032 but I have to admit that I nearly understood nothing when to use which identifier. The EdDSA identifier seems to be used for the family, whereas the other two are for the concrete incarnations. The Ed25519 and Ed448 identifiers are known both to BouncyCastle and to newer Java implementations (according to the related documentation), so I felt save to use them. What I have not tested is what Java returns as an algorithm for the keys. This is something I really could do.

demonti:
I have tested the KeyPairGenerator of Java 15 and it indeed returns keys marked with the algorithm "EdDSA", even if invoked with the "Ed25519" or "Ed448" algorithms. So there is a different behaviour between Java's built-in implementation and BouncyCastle.I therefore adjusted the "checkAlgorithm" method to accept both identifiers.

ibauersachs:
Thanks! Sorry for the delay, I wanted to check this against dnssecjava.

Issue #198, opened 2021-06-10
IDN / punycode support


Kent1:
Hello,

I've recently noticed that when using our dns tool (which uses dnsjava), we always got nxdomain for Internationalized Domain Names. I've seen later that the DNS request contained some unicode characters.

Is it by design that domains are not translated to the ASCII version or am i using the API wrongly perhaps ?

Thanks,
Quentin

ibauersachs:
Well, I guess you could call it by design, but it's more like nobody has yet gotten around to implement IDN translation in the `Name` class.

Kent1:
Ok, i'll make a pull request in that case :) Seems fairly easy, will look at your codebase.

Kent1:
I must say i am a bit confused by all the code in Name. Here some questions:

My POV is that the Punycode conversion could happen during the toString() call. However, looping over byte[] is not really best to apply the IDN algorithm. Is there a reason to story byte[] instead of String ?
However, doing the conversion only during toString() is not enough, since we should check length of labels not being greater than 63 when Punycode encoded.

Also, i see that Name escapes non-printable ASCII characters. Should that be replaced (partially) by IDN ? Partially since I also see characters like parenthesis or comma. Are those somewhere supported in domain names ?

Quentin

ibauersachs:
> I must say i am a bit confused by all the code in Name. Here some questions:
> 
> My POV is that the Punycode conversion could happen during the toString() call. However, looping over byte[] is not really best to apply the IDN algorithm. Is there a reason to story byte[] instead of String ?

The Name class is a representation of the DNS name, which is essentially a binary format. Storing it as such is IMO correct.

> However, doing the conversion only during toString() is not enough, since we should check length of labels not being greater than 63 when Punycode encoded.

You'll probably want to convert from e.g. `fenêtre.example.org` to `xn--fentre-kva.example.org` in Name constructor. `toString()` should then return the punycode, while e.g. `toUnicodeString()` (or an overload of `toString()`) could return the unicode string again. Storing the punycode encoded data in Name would definitely fix your NXDOMAIN issues.

> Also, i see that Name escapes non-printable ASCII characters. Should that be replaced (partially) by IDN ? Partially since I also see characters like parenthesis or comma. Are those somewhere supported in domain names ?

Not sure about the invalid characters, I'll need to look at that. AFAIK the Name constructor (or `fromString(...)`) should throw if it encounters invalid input.

ibauersachs:
[...]

PR #201, opened 2021-06-28
IDN support

Kent1:
Issue #198 

First draft. There is still a test failing where input contains escaped characters.

### DIFF
--- a/src/main/java/org/xbill/DNS/Name.java
+++ b/src/main/java/org/xbill/DNS/Name.java
@@ -206,9 +206,8 @@ public class Name implements Comparable<Name>, Serializable {
         return;
     }
     // Punycode encoding
-    String s1;
     try {
-      s1 = IDN.toASCII(s);
+      s = IDN.toASCII(s);
     } catch (IllegalArgumentException e) {
       throw parseException(s, "cannot be encoded to Punycode");
     }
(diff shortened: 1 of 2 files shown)


src/main/java/org/xbill/DNS/Name.java
```diff
@@ -204,32 +205,39 @@ public Name(String s, Name origin) throws TextParseException {
         copy(root, this);
         return;
     }
+    // Punycode encoding
+    String s1;
+    try {
+      s1 = IDN.toASCII(s);
```
ibauersachs, line 211: Use `s = IDN.toASCII(s);` to avoid the many not really needed further changes

src/test/java/org/xbill/DNS/NameTest.java
```diff
@@ -369,7 +369,7 @@ void ctor_toobig_escaped_end() {
       assertThrows(TextParseException.class, () -> new Name("ab\\256"));
     }
 
-    @Test
+//    @Test
```
ibauersachs, line 372: Reenable this test

src/test/java/org/xbill/DNS/NameTest.java
```diff
@@ -1249,6 +1249,14 @@ void toString_special_char() throws WireParseException {
     assertEquals(exp, n.toString());
   }
 
+  @Test
+  void toString_idn() throws TextParseException {
+    String in = "café.be";
```
ibauersachs, line 1254: please use a subdomain of example.com/net/org to avoid any possible confusion with a real domain

src/test/java/org/xbill/DNS/NameTest.java
```diff
@@ -1455,6 +1463,15 @@ void weird() throws TextParseException {
       assertNotEquals(n1, n2);
       assertNotEquals(n2, n1);
     }
+
+    @Test
+    void idn() throws TextParseException {
+      Name n1 = new Name("xn--caf-dma.be.");
+      Name n2 = new Name("café.be.");
```
ibauersachs, line 1470: same as in the other test, please use a example.com/net/org subdomain

ibauersachs (changes requested):
Thanks! Looks okay in general, but please see the inline comments. And you'll need to format the code with the formatter to make all the test run (`mvn com.coveo:fmt-maven-plugin:format`).

src/test/java/org/xbill/DNS/NameTest.java
Kent1: Well i first need to fix it. But not sure how. IDN does not support this escape characters. What would you advise ?

Kent1:
`com.coveo:fmt-maven-plugin:format` does not change anything.

src/test/java/org/xbill/DNS/NameTest.java
ibauersachs: Not sure it works, but could you un-escape them before passing it through the IDN conversion?

src/test/java/org/xbill/DNS/NameTest.java
```diff
@@ -1249,6 +1249,14 @@ void toString_special_char() throws WireParseException {
     assertEquals(exp, n.toString());
   }
 
+  @Test
+  void toString_idn() throws TextParseException {
```
hyperxpro, line 1253: Please use camel-case because this is Java.

hyperxpro (changes requested):

ibauersachs:
@Kent1 I'm working on this, i.e. running IDN'ization on a per-label basis.

PR #202, opened 2021-07-01
Rename echconfig to ech in SVCB/HTTPS records

adam-stoler:
In the latest SVCB RFC draft (https://datatracker.ietf.org/doc/html/draft-ietf-dnsop-svcb-https-06), the parameter "echconfig" has been renamed to "ech".  This Pull Request implements that change for dnsjava, and renames the classes and variables appropriately.

### DIFF
--- a/src/main/java/org/xbill/DNS/HTTPSRecord.java
+++ b/src/main/java/org/xbill/DNS/HTTPSRecord.java
@@ -7,7 +7,8 @@ import java.util.List;
  * HTTPS Service Location and Parameter Binding Record
  *
  * @see <a
- *     href="https://tools.ietf.org/html/draft-ietf-dnsop-svcb-https-01">draft-ietf-dnsop-svcb-https</a>
+ *     href="https://tools.ietf.org/html/draft-ietf-dnsop-svcb-https-06">draft-ietf-dnsop-svcb-https</a>
+ * @since 3.3
  */
 public class HTTPSRecord extends SVCBBase {
   HTTPSRecord() {}
--- a/src/main/java/org/xbill/DNS/SVCBBase.java
+++ b/src/main/java/org/xbill/DNS/SVCBBase.java
@@ -115,6 +115,8 @@ public abstract class SVCBBase extends Record {
     parameters.add(IPV4HINT, "ipv4hint", ParameterIpv4Hint::new);
     parameters.add(ECH, "ech", ParameterEch::new);
     parameters.add(IPV6HINT, "ipv6hint", ParameterIpv6Hint::new);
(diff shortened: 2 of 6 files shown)


ibauersachs (changes requested):
The changes break semantic versioning if I were to apply them without going to v4, and a rename is not significant enough for that.
We could either duplicate the `ParameterEchConfig` class and deprecate the existing, or keep the class name and only add `public static final int ECH = 5;` (but also keeping `ECHCONFIG` as deprecated).

adam-stoler:
Good point.  I was thinking that these were internal classes and values, but they are public so must remain consistent in the API.  I went the approach of duplicating and deprecating the old ones.  Let me know if you think that anything else is needed.

ibauersachs:
Actually yes, could you please add Javadoc with `@deprecated` tags mentioning what to use instead?
I also added `@since` tags to all new stuff added in or after 3.0, so if you wouldn't mind...

adam-stoler:
Sure, I added the javadoc @deprecated tags with links to the correct ones to use and added @since tags to the SVCB and HTTPS record classes.  I also made SVCBRecordBase public so it and its nested classes would show up in javadoc.

ibauersachs:
Thanks!
I was looking at this from my mobile so far and missed that e.g. `1 h3pool. alpn=h2,h3 echconfig=1234` is now not parseable anymore AFAICT. My gut feeling however says that we should be able to since this is what was produced until this patch. But you may convince me otherwise.

adam-stoler:
I was torn on this issue.  I generally think it is a good idea to maintain backwards compatibility, but since this is still not even an official RFC, I was not sure it was worth the effort to support the obsolete name.  Though now that I see how easy this is to support with the alias functionality in the Mnemonic class, it seems worthwhile to support this still just in case.

Note that with the implemented solution, it will always write out the RDATA with the "ech" name even if it was parsed from a string that used the obsolete "echconfig" name.  I think that is desirable (and simpler to implement), but let me know if you disagree.

ibauersachs:
> I was torn on this issue. I generally think it is a good idea to maintain backwards compatibility, but since this is still not even an official RFC, I was not sure it was worth the effort to support the obsolete name. Though now that I see how easy this is to support with the alias functionality in the Mnemonic class, it seems worthwhile to support this still just in case.

It's probably something that can be removed in v4 if the RFC is final.

> Note that with the implemented solution, it will always write out the RDATA with the "ech" name even if it was parsed from a string that used the obsolete "echconfig" name. I think that is desirable (and simpler to implement), but let me know if you disagree.

IMO that's the preferred way.

adam-stoler:
Sounds great, thanks!

PR #204, opened 2021-07-09
Fix issue with Name compareTo and the handling of octal digit byte values

adam-stoler:
The Name.compareTo() function has a bug with its handling of octal digits byte values that are greater than 128.  The lookup in the lowercase array was returning a signed byte that was implicitly cast to a signed int, resulting in a negative number.  This would cause such values to be incorrectly treated as less than those in the 0-127 range.

This PR fixes the issue and adds test to verify the behavior.  The fix is simply to perform another "& 0xFF" on the value that comes back from the lookup in the lowercase array.  The octal_digits_high() test would previously have failed without this fix.

### DIFF
--- a/src/main/java/org/xbill/DNS/Name.java
+++ b/src/main/java/org/xbill/DNS/Name.java
@@ -839,7 +839,9 @@ public class Name implements Comparable<Name>, Serializable {
       int length = name[start];
       int alength = arg.name[astart];
       for (int j = 0; j < length && j < alength; j++) {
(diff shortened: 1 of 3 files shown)


ibauersachs:
Can you please rebase this on master so that your svcb changes are not part of the PR?

While you're at int promotions, would you mind checking Name.hashCode() too? Sonar reports a bug there which I never got around to look at:
https://sonarcloud.io/project/issues?id=dnsjava_dnsjava

adam-stoler:
I am just going to close this PR and start a clean branch from master with the changes.
I will look into that SonarCloud issue as well before filing the new PR.

PR #207, opened 2021-08-03
IDN support for Name

ibauersachs:


ibauersachs:
@demonti AFAIK you're kind of a power user of dnsjava. Do you foresee any issues when IDN parsing is enabled by default in `Name`?

demonti:
sorry, I missed your question 10 days ago. While in general I am not against it, the first problem is that two versions of IDN exist in the real world, IDNA2003 and IDNA2008. The main difference for the perspective of the Germans is the handling of the German double s. While the old version converts the double s into two s, the new version keeps it and treats it like any other non-LDH character. However, there have also other issues been solved that have occurred with the old standard. Unfortunately, Java (including 16) still relies on the IDNA2003 standard, while TLD registries mostly have switched to IDNA2008, including, but not limited to, the German ccTLD DENIC. In our code, we use the respective functionality of the ICU library for the conversion forth and back, which is IDNA2008 compliant. So from this perspective, it is somewhat problematic. 

Another point is the potential processing overhead. In earlier times we used a library of a TLD registry and it turned out to be really slow. I have not done any performance tests on the built-in IDN class, however. But it is clear that various tests need to be done, even if the name turns out to be a plain LDH domain name.

Third, to me the impact on the whole DNS library is not really clear. Thinking of the Master class, the question is whether developers would then expect that zone files may contain IDNs at any places where domain names occur (both on the left and right of the record type). There would also be an encoding issue with the constructors that use InputStream as the source for the content – the library would have to assume a specific encoding or the API would have to be extended in a certain way.

So if the functionality is desired, my suggestions are:
* document in Javadoc that the functionality is based on java.net.IDN and uses IDNA2003
* maybe move the functionality to an alternative factory method "fromIDNString" and leave the constructor unchanged

ibauersachs:
Thank you! That is extremely valuable input and I'll probably revert the the default-enabled IDN parsing and go with your second suggestion, i.e. only parse IDN from a dedicated method, possibly with an option to choose the IDNA version (starting with only the built-in 2003-version).

I hadn't thought enough about e.g. use cases with zone files. I guess it's really how one is using this library: dealing with user input vs. with machine input.

PR #209, opened 2021-08-15
Import dnssecjava

### LINKED ISSUE #172: Merge dnssecjava into dnsjava
I was thinking for a while already to merge the DNSSEC resolver implemented in [dnssecjava](https://github.com/ibauersachs/dnssecjava) into dnsjava but wasn't sure this is the right thing to do.

Pros:
- dnssecjava is useless without dnsjava and closely depends on the API
- might proliferate DNSSEC a bit
- large test coverage
- parts of dnssecjava's tests depend on internal/private dnsjava API to achieve coverage

Cons:
- my personal interest
- different project licenses <sup>1</sup>
- change of package for dnssecjava
- jar size increase <sup>2</sup>

**Thoughts?**

<hr/>

1) The original license of dnssecjava was BSD-3-Clause (c) VeriSign. I applied LGPL and later EPL to my work, keeping the BSD-3-Clause notice to stay compliant with it. The test replay files from unbound are also BSD-3-Clause. The only external contributions were very small and by people privately asking for a more liberal license (when it was still LGPL), one from VeriSign. Relicensing it back to BSD-3-Clause is thus possible.
However, it would probably be best to keep a single license on dnsjava, meaning a switch from BSD-2-Clause to BSD-3-Clause (which is adding: `3. The name of the author may not be used to endorse or promote products derived from this software without specific prior written permission.`). I don't think this is an issue.

2) Size comparison:

Project | Compressed | Uncompressed
--- | --- | ---
dnsjava | 395 KiB | 811 KiB
dnssecjava | 51 KiB | 113 KiB
merged (including metadata) | 491 KiB | 923 KiB

ibauersachs:
Closes #172


src/main/java/org/xbill/DNS/dnssec/ByteArrayComparator.java
```diff
@@ -0,0 +1,34 @@
...
+
+import java.util.Comparator;
+
+/**
+ * This class implements a basic comparator for byte arrays. It is primarily useful for comparing
+ * RDATA portions of DNS records in doing DNSSEC canonical ordering.
+ *
+ * @since 3.5
+ */
+final class ByteArrayComparator implements Comparator<Object> {
```
hyperxpro, line 14: How about making this class Singleton?

ibauersachs:
@nresare I think this is ready for merging, can you have a look please? I'm mainly interested in comments about the API, the entire content is probably too much.

nresare (approved):
This is obviously a very substantial diff that it is difficult to review all aspects of, but I had a brief look and it seems to me like a solid improvement to dnsjava. Let's land this, and we can come back to any outstanding issues later :)

Issue #215, opened 2021-09-01
Unnecessary synchronization in org.xbill.DNS.Header::getID


maltalex:
I'm working on improving the performance of a project that deals with enormous amounts of dns messages. One of the performance bottlenecks I'm seeing is a whole lot of locking in `org.xbill.DNS.Header::getID` which is implemented like this:

```
private static final Random random = new SecureRandom();

public int getID() {
  synchronized (random) {
    if (id < 0) {
      id = random.nextInt(0xffff);
    }
    return id;
  }
}
```

The implementation is not only synchronized - it's synchronized around a `static` object, meaning that all `getID` calls in all `Header` instances are serial! Note that `SecureRandom` itself is thread-safe and doesn't require locking.

Could the `getID` implementation be changed to be lock-free using a `volatile` variable or an `AtomicInteger`? Or at least switched to use per-instance locking (e.g. synchronize on `this` or some other non-null field) to reduce the lock's scope?

I'd be happy to submit a PR with a lock-free implementation.

ibauersachs:
An `AtomicInteger` would generate predictable results, which is a no-go to avoid the Kaminsky attacks. Since SecureRandom is thread safe, the synchronization around it can be removed. And volatile would can be avoided by always generating an id (instead of -1) in the no-arg constructor.

If you open a PR, you may want to create it against the release/3.4.x branch as master won't get a new release soon.

ibauersachs:
> I was suggesting using an `AtomicInteger` not for id generation, but for storing the id and setting it using an atomic, lock-free `compareAndSet`.

Ah, I see.

> > And volatile would can be avoided by always generating an id (instead of -1) in the no-arg constructor.
> 
> That would be the best option IMHO. But but then we lose the lazy initialization of the ID field, which looks like the intention of the author of that class. Using `SecureRandom` too much can also be extremely slow (when low on entropy), so if there are use cases that generate many `Header` instances but don't need to initialize their ID field, they'd see worse performance due to this change. Is that a common scenario?

The default implementation of SecureRandom is SHA1PRNG, which AFAIK doesn't do reseeds and thus will never run out of entropy. I wouldn't know when an unset ID would be useful. Answers coming from the network have an ID which is parsed, and sending request need an ID anyway. The only case where the ID must be statically `0` is DNS over HTTP since the protocol is TCP and the order of delivery is guaranteed.

bwelling:
[...]

PR #216, opened 2021-09-02
removed synchronization in Header::getID

maltalex:
Header::getID synchronized on a static field causing unnecessary locking
especially when the ID field didn't need to be generated. As per the
discussion in #215, this was due to a desire to have the ID generation
be lazy, which is no longer needed.

This commit changes the ID generation behavior so the ID is either
explicitly supplied to the constructor, or generated by the constructor
using the same SecureRandom instance. This makes the getID method a
simple getter with no need for synchronization

### DIFF
--- a/src/main/java/org/xbill/DNS/Header.java
+++ b/src/main/java/org/xbill/DNS/Header.java
@@ -31,15 +31,14 @@ public class Header implements Cloneable {
    * @param id The message id
    */
   public Header(int id) {
-    this();
-    setID(id);
+    counts = new int[4];
+    flags = 0;
+    this.id = id;
   }
 
   /** Create a new empty header with a random message id */
   public Header() {
-    counts = new int[4];
-    flags = 0;
-    id = -1;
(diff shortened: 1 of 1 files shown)


src/main/java/org/xbill/DNS/Header.java
```diff
@@ -31,15 +31,14 @@
    * @param id The message id
    */
   public Header(int id) {
-    this();
-    setID(id);
+    counts = new int[4];
+    flags = 0;
+    this.id = id;
```
ibauersachs, line 36: This now allows out-of-range IDs. Since calling setID from a constructor isn't really a good idea (virtual member call), please either copy the range check or extract it into a private method.

ibauersachs (changes requested):

src/main/java/org/xbill/DNS/Header.java
maltalex: sure thing!

Issue #218, opened 2021-09-23
TCPClient incorrect receive timeout


nguydavi:
Hi,

I noticed that the TCPClient uses `endTime` as nanoseconds but this line still compares it with `currentTimeMillis`
https://github.com/dnsjava/dnsjava/blob/master/src/main/java/org/xbill/DNS/TCPClient.java#L115. This essentially makes the receive timeout being ignored.

This was introduced in https://github.com/dnsjava/dnsjava/commit/629f2eb3f5c5ff0695f8264dde03ed7666744775, and I don't know if there are other similar bugs in the other files that were changed.

Thanks!

ivanr:
We're running into this issue a lot in production. Are there any plans to make a new release that will include the fix? Thanks.

ibauersachs:
@ivanr I missed that this was only fixed in master and not in the 3.4.x branch. It would need a backport for a release, 3.5 (i.e. master) won't happen anymore this year.

ivanr:
@ibauersachs Could you please reopen this ticket until a fix is released?

PR #219, opened 2021-09-23
TCPClient: Fix recv timeout to properly use nanoseconds (#218)

nguydavi:


ibauersachs:
Thanks!
Actually, instead of just replace millis with nanos, the class should store the timeout value instead of endTime to avoid the numeric overflow mentioned in the Javadoc of nanoTime(). Would you mind refactoring TCPClient like this?
Also, 3.5 (current master) will take a while before there's a release. You might want to target the 3.4.x branch instead.

PR #221, opened 2021-10-01
Allow TSIG to accept well-known algorithm names

schtever:
The issue appears to be that the algorithm Name to java Mac name mapping is also being used for the reverse.  Not surprisingly, the java names for Mac algorithms do not match the well known names used by BIND and others.

This pull request separates out the String->Name mapping from the Name->Mac name mapping.  A few variants are added to allow some flexibility to the caller.  For example, they can specify the algorithm name with our without the trailing period.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,4 +5,4 @@
 *.iml
 target/
 .DS_Store
-
+.vscode/
--- a/src/main/java/org/xbill/DNS/TSIG.java
+++ b/src/main/java/org/xbill/DNS/TSIG.java
@@ -71,12 +71,42 @@ public class TSIG {
     algMap = Collections.unmodifiableMap(out);
   }
 
+  private static final Map<String, Name> stringToNameAlgMap;
+
+  static {
+    Map<String, Name> outsn = new HashMap<>();
+    outsn.put(HMAC_MD5   .toString()    .toLowerCase(), HMAC_MD5);
+    outsn.put(HMAC_MD5   .toString(true).toLowerCase(), HMAC_MD5);
+    outsn.put("hmac-md5"                .toLowerCase(), HMAC_MD5);
+    outsn.put("hmac-md5."               .toLowerCase(), HMAC_MD5);
+    outsn.put(HMAC_SHA1  .toString()    .toLowerCase(), HMAC_SHA1);
+    outsn.put(HMAC_SHA1  .toString(true).toLowerCase(), HMAC_SHA1);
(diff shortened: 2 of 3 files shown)


ibauersachs (changes requested):
Hi and thanks for your contribution!

The additional map you created is not needed, `algorithmToName` can be modified to achieve the same by not only comparing `alg` against the JCE algorithm name, but also against the key (with and without final dot).

The current state also includes `hmac-md5`, which is a completely undefined algorithm name, except for the incorrect Javadoc. Instead of restoring legacy behavior, the Javadoc should be updated to refer to the correct (i.e. RFC defined) name, including the dot, as well as mentioning that JCE HMAC names are also supported.

In light of [RFC8945](https://www.rfc-editor.org/rfc/rfc8945.html#section-5.2.2.1) however, I don't think JCE names should be supported at all and `nameToAlgorithm` marked as deprecated (so it can be made private in 4.x).

Thoughts?

schtever:
Hi and thanks for the feedback!

I agree that the second map is not needed if we are going to just allow the formally defined algorithm names.  In that case, then there's also no need for the `TSIG` constructor with the String for the algorithm.  The algorithm Names are public so callers could just use those.

My intent with the current approach was to provide a more forgiving String->Name transformation.  For example, with HMAC-MD5, that name is still in common use in many configurations.  ISC BIND, ISC DHCP, ISC KEA, PowerDNS, and Cisco Network Registrar all use the string "HMAC-MD5" to specify that algorithm in their configuration files.  Hence, I was hoping to honor that common usage by allowing the caller to specify "HMAC-MD5" instead of the the fully qualified RFC name.

I hesitated about continuing the support for the Java names, but thought that backwards compatibility should be preserved.

In the long run, if the goal is to only allow the RFC8945 names, then I'll reuse the existing map, eliminate the java names,  and deprecate `nameToAlgorithm`.  Should I also deprecate the `TSIG` constructor with the String algorithm argument?

ibauersachs:
Okay, let's keep the hmac-md5 name if it is also used by others, I wasn't aware of this. I'd still like to avoid the additional map, hmac-md5 can be treated as an exception. Also the Javadoc needs to be updated to indicate the support for RFC names and for the md5-exception.
We can continue to support the Java HMAC names for now, but that shouldn't be mentioned in the Javadoc and the nameToAlgorithm method marked as deprecated (both with the annotation and the Javadoc that explains why).

PR #237, opened 2021-11-22
Fix NPE when calling refresh before getting a config at all

### LINKED ISSUE #234: ResolverConfig.refresh() does not perform checkInitialized() ?
From the documentation of AndroidResolverConfigProvider I concluded that it could be a good approach:
- to call AndroidResolverConfigProvider.setContext() once in the begin of my Activity and
- to call ResolverConfig.refresh() each time before performing a lookup (with lookups being triggered manually via a Button).

Doing so I got a NullPointerException:
```
java.lang.NullPointerException: Attempt to invoke interface method 'java.util.Iterator java.util.List.iterator()' on a null object reference
 at org.xbill.DNS.ResolverConfig.<init>(ResolverConfig.java:106)
 at org.xbill.DNS.ResolverConfig.refresh(ResolverConfig.java:99)

```
... when calling refresh() for the first time because the list of configProviders hasn't been initialized, yet.
If my handler continues to perform the lookup anyway, the list of configProviders is initialized during the lookup.

Afterwards calling ResolverConfig.refresh() again does not throw a NullPointerException as the list of configProviders is now initialized.

**What is the correct way of dealing with that?**

Option 1:
I could call ResolverConfig.getCurrentConfig() once right after setting the context, because this also initializes the list of configProviders. But actually I do not need that information.
This way I could keep my Handler unchanged, i.e. calling refresh() right before each lookup (being triggered manually).

Option 2:
I could prevent a call to refresh() if there hasn't been a lookup before. I would have to memorize and check that within the Handler.

**Alternatively,** wouldn't it be better to call checkInitialized() in ResolverConfig.refresh()?
Or maybe within the ResolverConfig() constructor before looping over the list of configProviders?

In general:
Would you consider adding an example for use on Android in the documentation?

By the way: 
I saw with DNSJava v3.4.2, that refresh() correctly collects the name servers of the new mobile connection, but keeps the outdated name servers of the previous mobile connection.
This is already solved on the master with the added call to reset() in AndroidResolverConfigProvider.initialize(), right?

ibauersachs:
Closes #234

### DIFF
--- a/src/main/java/org/xbill/DNS/ResolverConfig.java
+++ b/src/main/java/org/xbill/DNS/ResolverConfig.java
@@ -55,20 +55,7 @@ public final class ResolverConfig {
   private static List<ResolverConfigProvider> configProviders;
 
   private static void checkInitialized() {
-    if (configProviders == null) {
-      configProviders = new ArrayList<>(8);
-      if (!Boolean.getBoolean(CONFIGPROVIDER_SKIP_INIT)) {
-        configProviders.add(new PropertyResolverConfigProvider());
-        configProviders.add(new ResolvConfResolverConfigProvider());
-        configProviders.add(new WindowsResolverConfigProvider());
-        configProviders.add(new AndroidResolverConfigProvider());
-        configProviders.add(new JndiContextResolverConfigProvider());
-        configProviders.add(new SunJvmResolverConfigProvider());
-        configProviders.add(new FallbackPropertyResolverConfigProvider());
-      }
-    }
-
-    if (currentConfig == null) {
+    if (currentConfig == null || configProviders == null) {
       refresh();
     }
   }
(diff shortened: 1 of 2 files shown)


nresare (approved):
This seems fine to me. One could argue that making refresh always succeed would make this a tiny bit more straight forward, but I have no strong opinion about this. Approved.

ibauersachs:
> This seems fine to me. One could argue that making refresh always succeed would make this a tiny bit more straight forward, but I have no strong opinion about this. Approved.

I'm not sure what you mean. Do you see a case where it still fails?

nresare:
I just meant that if we want `ResolverConfig.refresh()` to always succeed, even when configProviders have not yet been set, we can do that. That would be less fragile than making sure that refresh() is only called after checkInitialized() is called. 

My preferred solution would be to move the init code from checkInitialized() to the constructor, as that would man that partially initialised objects (where configProviders are not yet set) would never exist.

ibauersachs:
> I just meant that if we want `ResolverConfig.refresh()` to always succeed, even when configProviders have not yet been set, we can do that. That would be less fragile than making sure that refresh() is only called after checkInitialized() is called.
> 
> My preferred solution would be to move the init code from checkInitialized() to the constructor, as that would man that partially initialised objects (where configProviders are not yet set) would never exist.

`configProviders` is static, so I cannot completely avoid having `checkInitialized()` in `getCurrentConfig()` and `getConfigProviders()`. But I can move the first part into the constructor, keeping only `currentConfig == null` to eventually call refresh.

Issue #114, opened 2020-07-07
Make org.xbill.DNS.Record class Serializable


ironluca:
It helps in many situations, in addition, per the code, all the internal state of the Record and its sub-classes are Serializable. Serializable would be useful in place where the record has to be propagated to network for different parts of the system.

ibauersachs:
There is a very well defined serialization of Records: it's called DNS.
I intentionally removed the Serializable interface because the SerialVersionUID was not kept up to date and because java-binary serialization of Records is pointless with two already existing format (zone file format / DNS wire protocol). Additionally with Record now having a public constructor, it's not guaranteed to work anymore for all record types.

ironluca:
While I understand the above line of reasoning. Record being serializable would be useful in a few situations, as I am encountering. Just to highlight, I am using the records in a BPMN workflow context where the process state is serialized on encountering a User Task and reloaded for subsequent tasks. In this case, of course, the work around is to form a separate class with minimal information from the Record and make that Serializable.

However, as mentioned earlier, I appreciate why you don't want to make the class serializable, there are use cases where it would be actually very useful.

PR #242, opened 2021-12-26
Make Record serializable again

### LINKED ISSUE #132: org.xbill.DNS.Zone should not implement Serializable
`org.xbill.DNS.Record` of Branch master and 3.x does not implement `Serializable`. As a result `org.xbill.DNS.Zone` isn't serializable though it implements `Serializable`.

Note: `org.xbill.DNS.Record` of Branch 2.x does implement `Serializable`.

ibauersachs:
Instead of relying on the unstable default Java serialization of each
record type, this uses a proxy object. The proxy object only contains
the DNS wire data of a record, which is guaranteed to be stable. This
avoids all concerns about compatibility when Record classes evolve and
with inheritance.

Closes #132
See #114

### DIFF
--- a/src/main/java/org/xbill/DNS/Record.java
+++ b/src/main/java/org/xbill/DNS/Record.java
@@ -5,9 +5,14 @@ package org.xbill.DNS;
 
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
+import java.io.InvalidObjectException;
+import java.io.ObjectInputStream;
+import java.io.ObjectStreamException;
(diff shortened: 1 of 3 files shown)


nresare (approved):
We are fortunate in this project that we happen to have a very battle tested serialisation format for our value objects :)

As an aside, I don't think that the serialisation support in java was ever very well thought out to begin with, but that is not the fault of this change. It is simply magic that adding a non-public non-overriding `writeReplace()` method changes the serialisation behaviour.

PR #256, opened 2022-06-16
Ensure that empty TXT records always renders as ""

### LINKED ISSUE #254: Quoting of empty TXT records
Hi,

I'm not sure if this is a bug, but it is at least a different behaviour of creating empty TXT records while using the `TXTRecord` constructor or the static method `fromString` of `Record`:

```
        Record r0 = Record.fromString(recordName, Type.TXT, DClass.IN, 0, "", zoneName);
        TXTRecord r1 = new TXTRecord(recordName, DClass.IN, 0, "");
        System.out.println("r0: " + r0);
        System.out.println("r1: " + r1);
```
The output of this code snippet is:
```
r0: name.name.		0	IN	TXT
r1: name.name.		0	IN	TXT	""
```
BIND and PowerDNS servers are not able to load records with representation r0.

Patrick

nresare:
Instantiating an empty TXT record using Record.fromString() would
previously render as a TXT without the empty string.

Closes #254

### DIFF
--- a/src/main/java/org/xbill/DNS/TXTBase.java
+++ b/src/main/java/org/xbill/DNS/TXTBase.java
@@ -73,6 +73,10 @@ abstract class TXTBase extends Record {
   /** converts to a String */
   @Override
   protected String rrToString() {
+    if (strings.isEmpty()) {
+      // always return at least an empty quoted String
+      return "\"\"";
+    }
     StringBuilder sb = new StringBuilder();
     Iterator<byte[]> it = strings.iterator();
     while (it.hasNext()) {
--- a/src/test/java/org/xbill/DNS/RecordTest.java
+++ b/src/test/java/org/xbill/DNS/RecordTest.java
@@ -912,4 +912,12 @@ class RecordTest {
       }
     }
   }
+
+  // https://github.com/dnsjava/dnsjava/issues/254
+  @Test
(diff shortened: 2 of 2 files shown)


src/main/java/org/xbill/DNS/TXTBase.java
```diff
@@ -82,7 +82,9 @@ protected String rrToString() {
         sb.append(" ");
       }
     }
-    return sb.toString();
+    String s = sb.toString();
+    // always return at least an empty quoted String
+    return "".equals(s) ? "\"\"" : s;
```
ibauersachs, line 87: I'd rather do an `if (strings.isEmpty()){ return "\"\""; }` at the beginning to avoid unnecessary objects (iterator, string builder, string).

ibauersachs (changes requested):

ibauersachs (changes requested):
Oh, and please target this to the `release/3.5.x` branch.

src/main/java/org/xbill/DNS/TXTBase.java
nresare: Sure, that looks better. Updating.

nresare:
Updated according to feedback and re-targeted onto release/3.5.x

ibauersachs (approved):

PR #258, opened 2022-07-01
Minor cleanup

jkmcl:
- Convert an instance variable to local because its use is local
- Use String.replace instead of String.replaceAll when replacing a simple string (not regex)


jkmcl:
Incorrect change in 429e337

PR #259, opened 2022-07-01
Minor optimizations of SVCBBase

jkmcl:
- Use replace instead of replaceAll for better performance as "," is not a regex. Also, "\\\\," is equivalent to "\\," hence the latter is used
- Loop entry set of Map instead of key set when both key and value of each entry are used in every iteration of the loop to safe an extra call of Map.get


ibauersachs (approved):

ibauersachs:
Thanks!

PR #264, opened 2022-09-26
RRset as Lookup result to include RRSigs

sriram-km:
With Lookup, it isn't convenient to get RRSIGs. So a function named getRRset() to get the List<RRset>. Please look into it.

### DIFF
--- a/src/main/java/org/xbill/DNS/Lookup.java
+++ b/src/main/java/org/xbill/DNS/Lookup.java
@@ -63,6 +63,7 @@ public final class Lookup {
   private boolean doneCurrent;
   private List<Name> aliases;
   private Record[] answers;
+  private List<RRset> rrsets;
   private int result;
   private String error;
   private boolean nxdomain;
@@ -262,6 +263,7 @@ public final class Lookup {
     done = false;
     doneCurrent = false;
     aliases = null;
+    rrsets = null;
     answers = null;
     result = -1;
     error = null;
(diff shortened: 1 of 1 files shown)


ibauersachs:
Thanks for the idea. However, the proposed change only works in the happy case: a single answer to a query. But `Lookup` (and `LookupSession` too) follow CNAMEs, and in this case it would be unclear which `RRset`s are stored.
Also, the change doesn't consider the cache that is used for the second lookup of the same name and type.

Can you elaborate your use case?

sriram-km:
I am dealing with DNSSEC using dnsjava, I would like to get a record with its ```RRSIG```. Most developers use ```Lookup``` for looking up records hence having this will be better I believe, like getting the ```RRSIG``` of a record using ```Lookup```  and ```ValidatingResolver``` as Resolver. Now I am using ```ValidatingResolver``` and ```Record``` to query. I think getting ```RRset```s in ```Lookup``` will make things easy.

This is the use case. If my implementation has any issues. Could you please elaborate? I can improve them and fix them. 
Thanks

ibauersachs:
Sorry I didn't get back to you earlier. If the RRSigs were available, what would you be doing with them? An RRSig by itself doesn't provide much value.

What would you need in the non-happy path case, e.g. when a lookup fails (for any reason, e.g. DNSSEC validation, no record, no domain)? What about the RRSig of any intermediate CNAME records?

PR #269, opened 2022-11-27
fixing nullref exception

JoaoVictorVP:
A veeeery simple fix for a problem that occurs in Android 9.0

### DIFF
--- a/out/production/main/lombok.config
+++ b/out/production/main/lombok.config
@@ -1,0 +1,1 @@
+lombok.addLombokGeneratedAnnotation = true
--- a/src/main/java/org/xbill/DNS/Message.java
+++ b/src/main/java/org/xbill/DNS/Message.java
@@ -434,7 +434,8 @@ public class Message implements Cloneable {
         continue;
       }
       for (Record rec : sections[i]) {
-        rec.toWire(out, i, c);
+        if(rec != null)
+          rec.toWire(out, i, c);
       }
     }
   }


ibauersachs:
When does this happen? I'd rather prevent adding a `null` item to a section rather than just ignoring it. There are more places than just `toWire` where NPEs could occur.

JoaoVictorVP:
Next to this section of code, all I've done is create a message based on raw data from a udp packet. The code simply doesn't work in version 9.0 of android despite working perfectly in later versions, and it seems to me that, at least in this case, an NPE would only occur in that specific part.
image

ibauersachs:
How/where did you get the raw data?

And can you please post the raw data here (as base64 encoded)? I assume it's nothing confidential as your screenshot shows Youtube.

JoaoVictorVP:
The data is taken from the system itself through a VPN, processed with pcap4j and then I extract the packet data from these processed packets.

This issue only affects this specific version of Android, but the package's raw-data is (it actually happens with any package, but this is an example):
+M4BAAABAAAAAAAAD2luZmluaXRlZGF0YS1wYQpnb29nbGVhcGlzA2NvbQAAHAAB

ibauersachs:
> The data is taken from the system itself through a VPN, processed with pcap4j and then I extract the packet data from these processed packets.

Are you absolutely certain the capturing and pcap processing is correct?

> This issue only affects this specific version of Android, but the package's raw-data is (it actually happens with any package, but this is an example): +M4BAAABAAAAAAAAD2luZmluaXRlZGF0YS1wYQpnb29nbGVhcGlzA2NvbQAAHAAB

This message is a query for `nfinitedata-pa.googleapis.com./IN/AAAA` and decodes fine.

> (As an addendum, the fix I made legitimately fixed the problem in the project, so I think it's really just there)

I didn't find any code path within dnsjava that inserts `null` into a section. I thus want to find out how this actually happens.

PR #277, opened 2023-01-26
Fix NPE in SimpleResolver

PhroZenOne:
When reading a response that is REFUSED with no more data. The code could crash while comparing the DNS query with the response.

This change skips the comparison if the response is denied.

### DIFF
--- a/src/main/java/org/xbill/DNS/SimpleResolver.java
+++ b/src/main/java/org/xbill/DNS/SimpleResolver.java
@@ -402,6 +402,12 @@ public class SimpleResolver implements Resolver {
             return f;
           }
 
+          if (response.getQuestion() == null) {
+            f.completeExceptionally(
+                new WireParseException("invalid message: question section missing"));
+            return f;
+          }
+
           // validate name, class and type (rfc5452#section-9.1)
           if (!query.getQuestion().getName().equals(response.getQuestion().getName())) {
             f.completeExceptionally(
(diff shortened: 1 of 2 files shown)


ibauersachs:
Hi, thanks for your contribution!

I'm very hesitant to accept this as is, as I'm currently considering the answer you received from the server is invalid. Specifically, the DNS specification requires the validation on all of id/name/class/type (see [RFC 5452, 9.1](https://www.rfc-editor.org/rfc/rfc5452.html#section-9.1)). However, you're right that the missing name shouldn't cause a NPE: the response from `SimpleResolver` (and for that matter, the other implementations too) should be failed otherwise.

Can you please share some details about the setup where you got this, especially the name and version of the server?

PhroZenOne:
The DNS server is something I'm accessing as a third party so I have no info about the server itself. I have notified them and hope that they fix stuff on their side.

However, I changed the fix to throw an WireParseException instead. Only issue I have with this solution is that the DENIED flag might be a good indicator on what is wrong and that is hidden right now.

Maybe returning the message with the exception is a solution?

PhroZenOne:
@ibauersachs have you had time to think about this? :)

ibauersachs:
No, not yet. I'll might get to it on Sunday, but I can't make any promises.

ibauersachs:
It's fine like this, thank you!

PR #280, opened 2023-02-22
Configurable minimum selector timeout

vmarian2:
Add a system property to configure the minimum selector timeout

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -102,6 +102,12 @@ Do NOT use it.
 |true
 |false
 
+.2+|dnsjava.nio.selector_timeout
+3+|Set selector timeout. Default/Max 1000, Min 1.
+|Integer
+|1000
+|700
+
 4+h|dnssec options
 .2+|dnsjava.dnssec.keycache.max_ttl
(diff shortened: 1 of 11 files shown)


ibauersachs (changes requested):
Thanks for taking the time to address this.

- The `Options` class is somewhat legacy, I don't want to add more stuff there. Please use the property name `dnsjava.nio.selector_timeout` via `Integer.getInteger`.
- If the configured value is out of range (i.e. <= 0, larger than 1000), throw an exception.
- Document the new system property in the class' Javadoc and in the readme.
- Add a test if it's feasible, e.g. in NioTcpClientTest

If you want a release with this relatively soon, please target the branch `release/3.5.x` instead of master.

vmarian2:
I am unsure what exception to use in this case. Got any suggestions?

ibauersachs:
Please use IllegalArgumentException

vmarian2:
Closed in favour of https://github.com/dnsjava/dnsjava/pull/282 which is branched from release/3.5.x.

PR #282, opened 2023-03-14
Configurable minimum selector timeout

vmarian2:
As per the discussion in https://github.com/dnsjava/dnsjava/pull/280 just branched off release/3.5.x instead of master.


ibauersachs:
Thanks! I'll let this brew for a few days and will look at cutting a release next week.

PR #284, opened 2023-04-06
Refactored files to improve code quality and remove code smells

HetCMehta:
Refactored code in files ClientSubnetOption.java, dig.java, DNSOutput.java, ISDNRecord.java, Message.java, NSAPRecord.java, Record.java, ValidatingResolver.java, ValUtils.java, X25Record.java


ibauersachs:
Hi, thanks for taking time to work on dnsjava.
However, I had to stop reviewing your PR. If you want to continue working on this, please fix (at least) the following points.

- First of all, fix the formatting so that the diff is only about your actual changes.
- Don't change the public API.
- Not all record types have an address (regardless of the type), so those changes are wrong.
- Just because an IDE or some tool thinks a method is unsed, it actually isn't. This is a library, so everything that's public may be used by an application.

PR #289, opened 2023-07-28
Fix Name constructor failing with max length relative name and root origin

MMauro94:
The Name constructor failed when passing a max-length (253 chars) relative name and Name.root as origin, while this name is technically allowed.

The failure was caused by the Name constructor having the local "absolute" variable wrongly set to false in this case. This change fixes this by updating this flag when the origin is present.

Previous to this fix the workaround was to pass a null origin and manually append the final dot (root) to the string: in this case the constructor would work (as demonstrated by the "ctor_max_length_abs" test.

With this change I've also added a few tests for these edge-cases.

### DIFF
--- a/src/main/java/org/xbill/DNS/Name.java
+++ b/src/main/java/org/xbill/DNS/Name.java
@@ -271,6 +271,7 @@ public class Name implements Comparable<Name>, Serializable {
       appendFromString(s, label, pos);
     }
     if (origin != null && !absolute) {
+      absolute = origin.isAbsolute();
       appendFromString(s, origin.name, origin.labels);
     }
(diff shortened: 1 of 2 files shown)


ibauersachs:
Thanks!

The failed test is a bit brittle, I need to rework that. And the code coverage is probably lower because of the other failed test.

I'll have a deeper look at the changes, there was already an edge case about this somewhere.

ibauersachs:
Thanks again. Sorry for showing this as closed, I've cherry-picked the commit into the release/3.5.x branch.

MMauro94:
No worries, thank you :)

PR #294, opened 2023-10-12
BUG: Use executor in the `sendAsync(query, resolveListener)`

igr:
It starts with the `sendAsync(Message query, Executor executor)`. This method does not use `executor` at all.

Instead, it calls `sendAsync(Message, ResolverListener)`, which calls `sendAsync(query)` that is executed on common `ForkJoinPool`.

It seems we are missing a default `sendAsync(Message query, ResolverListener listener, Executor executor)`.

Wdyt?

EDIT: I forgot to mention that the existing code may lead to endless recursive calls; this is how I figured out the issue.

⚠️ Important: The inter-callilng of `Async` is _very_ messy. I am seeing the recursion loop between functions. Before my change, the loop consists of 3 functions. Now its 2, but still exist. It has to be fixed :(

### DIFF
--- a/src/main/java/org/xbill/DNS/Resolver.java
+++ b/src/main/java/org/xbill/DNS/Resolver.java
@@ -253,7 +253,7 @@ public interface Resolver {
 
           listener.receiveMessage(id, result);
           return null;
-        });
+        }, executor);
     return id;
   }
 }
--- a/src/main/java/org/xbill/DNS/Resolver.java
+++ b/src/main/java/org/xbill/DNS/Resolver.java
@@ -212,7 +212,7 @@ public interface Resolver {
           public void handleException(Object id, Exception e) {
             f.completeExceptionally(e);
           }
-        });
+        }, executor);
     return f;
   }
 
@@ -231,8 +231,13 @@ public interface Resolver {
    */
   @Deprecated
   default Object sendAsync(Message query, ResolverListener listener) {
+    return sendAsync(query, listener, ForkJoinPool.commonPool());
+  }
+
+  @Deprecated
+  default Object sendAsync(Message query, ResolverListener listener, Executor executor) {
     final Object id = new Object();
-    CompletionStage<Message> f = sendAsync(query);
+    CompletionStage<Message> f = sendAsync(query, executor);
     f.handleAsync(
         (result, throwable) -> {
           if (throwable != null) {


ibauersachs:
This is working as intended. The ResolverListener interface is deprecated and from pre-v3 times. If you read the Javadoc, it very clearly states the recursion if none of the methods are properly implemented.

igr:
> If you read the Javadoc, it very clearly states the recursion if none of the methods are properly implemented.

Please, the interface should be a rock-solid contract. There should not be a "one specific" way to implement it. The "must" part must also be specified using code, not Javadoc - you probably need to extract more methods in the interface.

As a result, I cannot adapt the library to use a custom thread pool: I also get the recursion with the existing implementation. 

Example:

```java
public class PooledResolverWrapper {
    public static final BoundedCachedThreadPool POOL = new BoundedCachedThreadPool(20);
    public static PooledResolverWrapper of(final Resolver resolver) {
        return new PooledResolverWrapper(POOL, resolver);
    }

    private final BoundedCachedThreadPool pool;
    private final Resolver resolver;
    public PooledResolverWrapper(final BoundedCachedThreadPool pool, final Resolver resolver) {
        this.pool = pool;
        this.resolver = resolver;
    }

    public CompletionStage<Message> sendAsync(final Message query) {
        return resolver.sendAsync(query, pool);
    }
}
```

If I apply this wrapper on any resolver implementation:

```java
public class PooledSimpleResolver extends SimpleResolver {
    private final PooledResolverWrapper wrapper = PooledResolverWrapper.of(this);
    @Override
    public CompletionStage<Message> sendAsync(final Message query) throws IOException {
        return wrapper.send(query);
    }
}
```

this _should_ work: we just delegated the execution of one method to the other one. However, there is a case when recursion happens.

ibauersachs:
I'm not quite sure what you were trying to achieve and what you're actual code looks like. However, the snippet below works fine.

The `Resolver` interface had a legacy method of doing async queries: `Object sendAsync(Message query, ResolverListener listener)`. This is deprecated in favor of real async with `CompletionStage`. To maintain backwards compatibility, the new method has a default implementation calling the old one.
To prevent deprecation warnings for users that migrated to the new method, the old one _also_ has a default implementation. This is only an issue if one works with the interface directly and implements neither method. Actual implementation like `SimpleResolver` override the new `CompletionStage`-based `sendAsync` and thus has no issue.

If there's another issue that I didn't catch, I'll have a look. But it's out of the question to introduce new methods with the `ResolverListener` interface or breaking backwards compatibility.

```java
import java.net.UnknownHostException;
import java.util.concurrent.CompletionStage;
import java.util.concurrent.ExecutorService;
import java.util.concurrent.Executors;
import org.xbill.DNS.DClass;
import org.xbill.DNS.Message;
import org.xbill.DNS.Name;
import org.xbill.DNS.Record;
import org.xbill.DNS.Resolver;
import org.xbill.DNS.SimpleResolver;
import org.xbill.DNS.Type;

public class Test {
  public static void main(String[] args) throws Exception {
    Message query =
        Message.newQuery(
            Record.newRecord(Name.fromConstantString("example.com."), Type.A, DClass.IN));

    PooledSimpleResolver x = new PooledSimpleResolver();
    System.out.println(x.sendAsync(query).toCompletableFuture().get());
  }

  public static class PooledSimpleResolver extends SimpleResolver {
    private final PooledResolverWrapper wrapper = PooledResolverWrapper.of(this);

    public PooledSimpleResolver() throws UnknownHostException {}

    @Override
    public CompletionStage<Message> sendAsync(final Message query) {
      return wrapper.sendAsync(query);
    }
  }

  public static class PooledResolverWrapper {
    public static final ExecutorService POOL =
        Executors.newFixedThreadPool(
            5,
            r -> {
              Thread t = Executors.defaultThreadFactory().newThread(r);
              t.setDaemon(true);
              return t;
            });

    public static PooledResolverWrapper of(final Resolver resolver) {
      return new PooledResolverWrapper(POOL, resolver);
    }

    private final ExecutorService pool;
    private final Resolver resolver;

    public PooledResolverWrapper(final ExecutorService pool, final Resolver resolver) {
      this.pool = pool;
      this.resolver = resolver;
    }

    public CompletionStage<Message> sendAsync(final Message query) {
      return resolver.sendAsync(query, pool);
    }
  }
}
```

igr:
Ignore my PR and code attempt, it's not ok.

anyhow, LSP is broken here, as behavior depends on the implementation. When you wrap SimpleResolver it works, but if you wrap a different implementation, it may lead to recursion. the solution would be to separate the thread scheduling from the actual behavior. no worries, I understand that isn't going to be fixed. I have copy/pasted and override a different sync method, that works with all resolver implementations we are using.

PR #296, opened 2023-10-23
Master issue295 tsig fix

frankarinnet:
This pull request is in reference to: https://github.com/dnsjava/dnsjava/issues/295

There are still some TODOs, however, because I need some guidance on how to deal with thread safety in the TSIG.

So, don't merge this without commenting.

Issue #295, opened 2023-10-18
bug with TSIG verification of large responses


frankarinnet:
We've encountered a bug with how DNSJAVA verifies the transaction signature of large responses.

In our environment, we can successfully perform an AXFR for a small zone (22 line zone file, 2096 bytes on the wire).  But we get a TSIG related error (org.xbill.DNS.ZoneTransferException: TSIG failure: BADSIG) when transferring a larger zone (7855 line zone file,  400k on the wire).

This bug was introduced between versions 3.1.0 and 3.20.  The TSIG class saw a significant rework there, and the interactions between the TSIG.StreamVerifier class and the TSIG.verify() method were broken then -- although seemingly only in cases where the signed response encompassed multiple messages.

While investigating this, we developed a patch for 3.5.2 this that we would be happy to submit as a pull request (or, really, help get this fixed some other way).

ibauersachs:
Thanks for the report. Since you already have a patch, opening a PR would be super helpful.

frankarinnet:
Willdo.  I'll get it submitted properly on Monday.

frankarinnet:
Pull request submitted, but I need some guidance on a few points.  I've marked them in the pull request with some TODOs.

I tested the code, though against responses of various sizes (1 message in length, 50some messages in length, and 300+ messages in length).

frankarinnet:
A few more notes.  The 3.2.0 version -- though bugged -- seemed to want HMAC signing logic moved out of the StreamVerifier class and into the TSIG.verify() method.

I have followed that in my changes, but it results in some extra conditional logic in the TSIG.verify() method.  Seeing the result, I'm a bit ambivalent.

If the maintainers decide they dislike the "everything goes in TSIG.verify()" method, I have a version which keeps more logic in the StreamVerifier class.

ibauersachs:
Thanks again! Would you mind just posting the alternative here or in a separate PR? It doesn't need to be pretty!

frankarinnet:
Pull request for alternate approach (https://github.com/dnsjava/dnsjava/pull/297) posted.

frankarinnet:
Added PR for main approach that properly targets 3.5.x branch: https://github.com/dnsjava/dnsjava/pull/298

(obseletes https://github.com/dnsjava/dnsjava/pull/296)

frankarinnet:
Added program (https://github.com/frankarinnet/axfr/blob/main/README.md) to help with manual testing.

[...]

PR #297, opened 2023-10-23
Issue 295, (alternate) addresses bug with TSIG verification of large responses

frankarinnet:
This is an alternate approach to fixing the large responses TSIG bug (https://github.com/dnsjava/dnsjava/issues/295) -- really only added for comparison purposes.

Like the other pull request (https://github.com/dnsjava/dnsjava/pull/296), do not accept without understanding and making some decisions about thread safety.


frankarinnet:
Added #301 which shows some necessary changes to #300 for adding the signature of the TSIG record at the right time.

PR #300, opened 2023-10-29
Fix handling of multi-message TSIG responses

### LINKED ISSUE #295: bug with TSIG verification of large responses

ibauersachs:
Closes #295
Closes #297
Closes #298
Closes #299
Closes #301


frankarinnet:
I'm afraid its not quite right.  In this case, there is still a bug with verify process (and a symmetric bug with the generate process).  The signature of the TSIG record isn't being added at quite the right time (as usual only in the case of mixed signed/unsigned messages).

On the bright side, the bug is much more narrow at this point.

I've got a branch of 'tsig-stream' I'll submit shortly as another PR.

src/main/java/org/xbill/DNS/TSIG.java
```diff
@@ -749,15 +849,16 @@ public int verify(Message m, byte[] b) {
...
-        lastTSIG = tsig;
+        int result = key.verify(m, b, queryTsig, true, sharedHmac);
+        hmacAddSignature(sharedHmac, tsig);
+        lastsigned = nresponses;
         return result;
       }
 
       if (tsig != null) {
-        int result = key.verify(m, b, lastTSIG, false);
+        int result = key.verify(m, b, null, false, sharedHmac);
```
frankarinnet, line 859: Very clever.  I think this satisfies all my concerns about backward compatibility.

Issue #253, opened 2022-05-16
Implement pluggable I/O?


ivanr:
As a workaround for #223 we maintain a limited fork of some dnsjava classes. We wanted to use our own NioTcpClient and NioUdpClient classes, but SimpleResolver calls their static methods so we had to fork that too. This broke down with dnsjava 3.5.1, which now calls Message#getGeneratedTSIG(), which is package-private.

For use cases like ours I think a cleaner approach would be to make the I/O pluggable so that we can use dnsjava directly while supplying our own I/O classes. If this is something you'd accept, is there an approach that you would like to take? Or should I propose something?

ivanr:
Would you be happy with an addition of a factory class that by default returns instances of the existing TCP and UDP client classes, but can be configured to return something else? Static calls to TCP and UDP clients would be replaced with static calls to this class instead.

ibauersachs:
That's seems okay. I wonder if using Java's `ServiceLoader.load` would make sense, but it's probably overkill.

ivanr:
I agree. I don't think the increased complexity of ServiceLoader brings anything. With the factory approach I may also try to support per-instance customisation. I will try to schedule some time for this soon.

ivanr:
@ibauersachs Just to let you know that I haven't dropped this. It's just that I've been extraordinarily busy. I will get to this eventually.

chrisruffalo:
I ran into this issue so I took a rough swing at it: https://github.com/dnsjava/dnsjava/pull/302

Obviously very rough. The main issue I ran into is the repeated use of package private on methods that may, in fact, form part of a more public API.

ibauersachs:
I tried to keep a lot of things package-private to avoid a broad API surface that cannot be changed later. Are there any particular methods you would like to see public?

chrisruffalo:
I think that mostly it was frustration in not being able to overwrite SimpleResolver but instead the need to copy it in it's entirety _and_ make it part of the org.xbill.DNS package which wouldn't work out cross-jar. I think making the resolver client factory (probably a bad name for it) helps ameliorate this issue a lot. I don't remember any methods in particular and I'm only part of the way into doing what I want. (More control over the internals of SimpleResolver while re-using the send/receive logic.)

PR #302, opened 2023-10-30
Allow end-user to construct clients for SimpleResolver

chrisruffalo:
- added ResolverClientFactory to allow creation of clients for the simple resolver
- added a default implementation (DefaultResolverClientFactory) that returns a flyweight implementation
- added a flyweight client implementation that is simply a pass-through to the existing static methods
- added factory to SimpleResolver class
  - used simple resolver during send/recv to construct correct client and send message
  - client is aware of if it is tcp or udp at construction time

This is very much a WIP pointed at issue #253

### DIFF
--- a/src/main/java/org/xbill/DNS/ResolverClientFactory.java
+++ b/src/main/java/org/xbill/DNS/ResolverClientFactory.java
@@ -1,0 +1,33 @@
+// SPDX-License-Identifier: BSD-3-Clause
+// Copyright (c) 2005 Brian Wellington (bwelling@xbill.org)
+
+package org.xbill.DNS;
+
+/**
+ * Interface for creating the TCP/UDP factories necessary for the simple resolver.
+ *
+ * @since 3.6
+ */
+public interface ResolverClientFactory {
+
+  /**
+   * Create or return a cached/reused instance of the TCP resolver that should be used to send UDP
+   * over the wire to the remote target. <br>
+   * It is the responsibility of this method to manage pooling or connection reuse. This method is
+   * called right before the connection is made every time the simple resolver is called. The
+   * implementer of this method should be aware and choose how to pool or reuse connections.
+   *
+   * @since 3.6
+   * @return an instance of the tcp resolver client
+   */
+  TcpResolverClient createOrGetTcpClient();
+
+  /**
+   * Create or return a cached/reused instance of the UDP resolver that should be used to send UDP
+   * over the wire to the remote target.
+   *
+   * @since 3.6
+   * @return an instance of the udp resolver client
+   */
+  UdpResolverClient createOrGetUdpClient();
+}
--- a/src/main/java/org/xbill/DNS/SimpleResolver.java
+++ b/src/main/java/org/xbill/DNS/SimpleResolver.java
@@ -14,6 +14,7 @@ import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionStage;
 import java.util.concurrent.Executor;
 import java.util.concurrent.ForkJoinPool;
+import lombok.Getter;
 import lombok.extern.slf4j.Slf4j;
 
 /**
@@ -43,6 +44,8 @@ public class SimpleResolver implements Resolver {
   private Duration timeoutValue = Duration.ofSeconds(10);
 
   private static final short DEFAULT_UDPSIZE = 512;
+
+  @Getter private ResolverClientFactory clientFactory = new DefaultResolverClientFactory();
 
   private static InetSocketAddress defaultResolver =
       new InetSocketAddress(InetAddress.getLoopbackAddress(), DEFAULT_PORT);
@@ -176,6 +179,16 @@ public class SimpleResolver implements Resolver {
   @Override
   public void setTCP(boolean flag) {
     this.useTCP = flag;
+  }
+
+  /**
+   * Allows external users to define the factory that creates clients for writing messages to the
+   * wire.
+   *
+   * @since 3.6
+   */
+  public void setClientFactory(ResolverClientFactory clientFactory) {
+    this.clientFactory = clientFactory;
   }
 
   /**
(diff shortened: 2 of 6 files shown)


src/main/java/org/xbill/DNS/FlyweightResolverClient.java
```diff
@@ -0,0 +1,18 @@
+package org.xbill.DNS;
+
+import java.net.InetSocketAddress;
+import java.time.Duration;
+import java.util.concurrent.CompletableFuture;
+
+public class FlyweightResolverClient implements TcpResolverClient, UdpResolverClient {
```
ibauersachs, line 7: Please name this `DefaultResolverClient`

src/main/java/org/xbill/DNS/SimpleResolver.java
```diff
@@ -44,6 +44,8 @@ public class SimpleResolver implements Resolver {
 
   private static final short DEFAULT_UDPSIZE = 512;
 
+  private ResolverClientFactory clientFactory = new DefaultResolverClientFactory();
```
ibauersachs, line 47: What's the intended way to change this from the default to something else?

src/main/java/org/xbill/DNS/TcpResolverClient.java
```diff
@@ -0,0 +1,18 @@
...
+
+import org.xbill.DNS.Message;
+
+import java.net.InetSocketAddress;
+import java.time.Duration;
+import java.util.concurrent.CompletableFuture;
+
+public interface TcpResolverClient {
+
+  CompletableFuture<byte[]> sendrecv(
```
ibauersachs, line 11: Please name this method `sendAndReceiveTcp`

src/main/java/org/xbill/DNS/UdpResolverClient.java
```diff
@@ -0,0 +1,19 @@
...
+
+import org.xbill.DNS.Message;
+
+import java.net.InetSocketAddress;
+import java.time.Duration;
+import java.util.concurrent.CompletableFuture;
+
+public interface UdpResolverClient {
+
+  CompletableFuture<byte[]> sendrecv(
```
ibauersachs, line 11: Please name this method `sendAndReceiveUdp`

src/main/java/org/xbill/DNS/ResolverClientFactory.java
```diff
@@ -0,0 +1,13 @@
+package org.xbill.DNS;
+
+/**
+ * Interface for creating the TCP/UDP factories
+ * necessary for the simple resolver.
+ */
+public interface ResolverClientFactory {
+
+  TcpResolverClient createTcpClient();
```
ibauersachs, line 9: Given that this might be an expensive operation, it's not clear that this method is called for every query. Maybe `getTcpClient()` would be more appropriate? OTOH `create` is more natural for a factory, but then the SimpleResolver should only call this once (e.g. in the constructor).

ibauersachs (changes requested):
Thank you very much! Happy to get this integrated.

Apart from the inline-comments, could you please address the following points in all files?
- SPDX license line
- Javadoc for every class and all public methods
- `@since 3.6` tags for new public classes and methods
- run `mvn fmt:format`

src/main/java/org/xbill/DNS/SimpleResolver.java
chrisruffalo: I forgot the setter.

src/main/java/org/xbill/DNS/ResolverClientFactory.java
chrisruffalo: I am not sure who should have the responsibility here and if it should be the same for UDP. My thought was that the factory would handle the pools and that create might be a bit of a misnomer. Is `createOrGet` overthinking it?

chrisruffalo:
I am working to resolve all the comments but I want to point out that this is very much a WIP so I'm not sure how much I want to invest in getting it just right if it is not viable/going in the right direction. Glad for the feedback though.

src/main/java/org/xbill/DNS/ResolverClientFactory.java
ibauersachs: createOrGet is fine.

ibauersachs:
> I am working to resolve all the comments but I want to point out that this is very much a WIP so I'm not sure how much I want to invest in getting it just right if it is not viable/going in the right direction. Glad for the feedback though.

Okay. Thanks anyway for getting this started and the updates. I'll leave this open for a while in case you find something that doesn't work quite right yet. I would assume that vert.x clients should work with this implementation (with Future <-> CompletableFuture conversions and an Executor that uses the event loop).

chrisruffalo:
I was able to get the vertx client working as a drop-in replacement through the interface. I'm not 100% happy with the semantics/naming and I almost wonder if the TCP and UDP factories should be entirely separate. Or perhaps use the same sendrecv method so that the factory choice determines the client in use?

I'm not sure what has the best "feel" though all the different permutations will work.

chrisruffalo:
I was also able to use the SSL engine in vertx to implement a DNS-over-TLS resolver that works with cloudflare. I am pretty happy with the progress I was able to make.

chrisruffalo:
I updated all of the comments and @since annotations

src/main/java/org/xbill/DNS/ResolverClientFactory.java
ibauersachs: That's fine.

ibauersachs:
I've worked with your branch, made some changes and rebased it on master, see here: [branch](https://github.com/dnsjava/dnsjava/tree/issue-253_ingo) or the [commit](https://github.com/chrisruffalo/dnsjava/commit/2c30c9065b06767879eeca8fbeb04119f6719e41).

One of the major changes was to replace `Resolver` with `Io` in the names, as these implementations are not necessarily restricted to the resolvers. What do you think?

ibauersachs:
Merged in master as 5da277075230e8a131edfefadfb119c0556fa904

PR #309, opened 2024-01-29
Remove unreachable code

hablutzel1:
Remove unreachable code to keep it simple as this is better for security code. Additionally, one simple typo is corrected for a private method.


ibauersachs:
Thanks!

Issue #334, opened 2024-08-20
bug in zone class wrt SOA/RRSig


frankarinnet:
The Zone class contructors throw an exception if you pass in a list of records that contains both a SOA and an RRSig for that SOA.

It appears as though the underlying Zone.addRecord() method is supposed to decorate the SOA record with its associated RRSig info.  But, due to this bug, it tries to replace the SOA record with the RRSig record (which fails).

See the attached stack trace: [stacktrace.txt](https://github.com/user-attachments/files/16667122/stacktrace.txt)

This seems to have been introduced with version "3.6.0".

frankarinnet:
The PR:
https://github.com/dnsjava/dnsjava/pull/335
is how I think this should be fixed.

It also includes a test case that demonstrates the bug.

frankarinnet:
You can also reproduce the bug by passing a signed zone to the "Zone.Zone(Name zone, String file)" constructor.
The following link:
  https://dnsinstitute.com/documentation/dnssec-guide/ch04.html
had a nice zone I used for some testing (near the text "As you can see, the zone grew in size, and the number of records multiplied.")

PR #335, opened 2024-08-20
fix bug in zone class wrt SOA/RRSig

frankarinnet:
Associated issue: https://github.com/dnsjava/dnsjava/issues/334

Its a few lines of changes to "Zone.addRecord()" that allow for ingesting zones that have both a SOA and an associated RRSig/SOA.

Additionally, there's a test case that demonstrates the bug (fails on 3.6.1, works if the fix is applied).

### DIFF
--- a/src/main/java/org/xbill/DNS/Zone.java
+++ b/src/main/java/org/xbill/DNS/Zone.java
@@ -234,6 +234,7 @@ public class Zone implements Serializable, Iterable<RRset> {
 
     Name name = r.getName();
     int rtype = r.getRRsetType();
+    int actualType = r.getType();
 
     if (rtype == Type.SOA && !name.equals(origin)) {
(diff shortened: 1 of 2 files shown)


ibauersachs:
Thanks a lot for noticing and coming up with a PR!

Did you also test if replacing a signed SOA still works? I wonder if the remembering only the SOA record alone (instead of its rrset) is even working in that case.

If the license allows it, we could maybe import the sample zone file.

frankarinnet:
re: replacing signed SOAs: I created a second test case for that.  It seemed to work, but you should look at it carefully.  I'm not 100% sure it addresses your question.

re: the sample zone file: it was just the one I was able to find browsing google.  You'd have to ask them ("The DNS Institute") about licensing.   The properly-signed examples I have access to are all reverse-related and anyway have our own contacts all over them.

PR #337, opened 2024-08-31
Add new IANA Trust Anchor

technoLord:
New TA is available and is expected to start signing in 2026:
https://data.iana.org/root-anchors/root-anchors.xml 

Worked on together with Poshan Peeroo.
( poshanpeeroo@cyberstorm.mu )


ibauersachs:
Thanks!

PR #345, opened 2024-10-07
Fix 'initialRequest' guard might be incorrect in DohResolver and potential 'rollback' on 'lastRequest' among concurrent requests.

LinZong:
Hi all, thanks for such solid & amazing dns implementation in Java world!
Trying to employ the DoH capability to my project, and found that there might be two hard-to-find issues lying in current [DoH implementation](https://github.com/dnsjava/dnsjava/blob/master/src/main/java/org/xbill/DNS/DohResolver.java).

1. 'initialRequest' guard might be incorrect if the initial value of 'idleConnectionTimeout' is set to a value larger than current (and later for a period) System.nanoTime(). 

Related code snippet:

https://github.com/dnsjava/dnsjava/blob/a6371af00867d13e5b3dbee040f640e93420fb21/src/main/java/org/xbill/DNS/DohResolver.java#L478-L483

Imaging that there is a service running DohResolver on a server booted recently, when DohResolver is about to send the first request, its `lastRequest.get() == 0L`, `idleConnectionTimeout.get() == some user-defined value from constructor or 2mins by default`. If `idleConnectionTimeout` is larger than current `System.nanoTime()`, the value of boolean guard `isInitialRequest` will be false, thing goes wrong.

(Eg: someone constructs a DohResolver with unary constructor but its server was booted 1min before. On my Ubuntu 18.04 server with Azul Zulu 8.0.322 JVM, `System.nanoTime()` is equals to my server's uptime.)
image

To address such case, we may need to introduce a new state variable to indicate the 'initial request' was acutally completed, to ensure the connection pool was created after the first connection was established successfully.

https://github.com/dnsjava/dnsjava/blob/e1e5c0c7ffb1e91dc604459a2855b39e9cc3c496/src/main/java/org/xbill/DNS/DohResolver.java#L478-L488

2. Timestamp variable 'lastRequest' might occur race condition on concurrent DoH request/response.

https://github.com/dnsjava/dnsjava/blob/a6371af00867d13e5b3dbee040f640e93420fb21/src/main/java/org/xbill/DNS/DohResolver.java#L541-L557

If there are two concurrent DoH requests happend, one with a smaller 'startTime' got the response from server behind of another one with a larger 'startTime', then the value 'lastRequest' will first be set to the large one, then 'bump' to the small one. 

We may also need to introduce a CAS get-and-set operation like code snippet below to ensure value in 'lastRequest' field is always monotonic.

https://github.com/dnsjava/dnsjava/blob/e1e5c0c7ffb1e91dc604459a2855b39e9cc3c496/src/main/java/org/xbill/DNS/DohResolver.java#L540-L552

### DIFF
--- a/src/main/java/org/xbill/DNS/DohResolver.java
+++ b/src/main/java/org/xbill/DNS/DohResolver.java
@@ -23,7 +23,6 @@ import java.util.concurrent.CompletionStage;
 import java.util.concurrent.Executor;
 import java.util.concurrent.ForkJoinPool;
 import java.util.concurrent.TimeUnit;
-import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.function.Function;
 import javax.net.ssl.HttpsURLConnection;
@@ -92,8 +91,6 @@ public final class DohResolver implements Resolver {
   private final AsyncSemaphore maxConcurrentRequests;
 
   private final AtomicLong lastRequest = new AtomicLong(0);
-
-  private final AtomicBoolean initialRequestSentMark = new AtomicBoolean(false);
   private final AsyncSemaphore initialRequestLock = new AsyncSemaphore(1);
 
   private static final String APPLICATION_DNS_MESSAGE = "application/dns-message";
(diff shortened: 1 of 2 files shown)


ibauersachs:
Thanks, I'll need some more time to properly review this.

ibauersachs:
Thanks for the detailed report and test!

The issue was the simple mistake of not calculating as mandated by the [System.nanoTime()](https://docs.oracle.com/javase/8/docs/api/java/lang/System.html#nanoTime--) documentation.

The test was flaky, I thus changed it to fake the provided time.

PR #354, opened 2024-11-29
Fixed confusing parameter name in CNAMERecord

chkal:
The `CNAMERecord` class has a constructor parameter for the CNAME target which is called `alias` at the moment. This is quite confusing, because:

- The API docs states that this parameter is "The name to which the CNAME alias points". So it is NOT the alias itself (which is the name of the record), but the target to which the record points.
- There is a deprecated method `CNAMERecord#getAlias()` which does NOT return the value supplied with the `alias` constructor parameter, but returns the name of the record instead.

I think it would be help to rename this parameter to `target`. That is also consistent with the name used in `NSRecord` and  `PTRRecord`.


ibauersachs:
Agree. Thanks for the cleanup!

PR #371, opened 2025-01-19
Hosts parser change

ibauersachs:


ibauersachs:
@ivanr-rs can you please verify if these changes work for you?
@tzaeschke is this enough log reduction?

tzaeschke:
Thanks for looking into this!
I tried it and the original warning has disappeared.
However, I now see a different warning:
`[main] WARN org.xbill.DNS.hosts.HostsFileParser - Failed to parse entire hosts file /etc/hosts, address failures=3, name failures=0`
It is an improvement in cases where there are multiple of our entries in the `/etc/hosts` file, otherwise it is just a different log message.
I would still prefer no warning at all, just like the JDK's or OS's own DNS resolver, if possible.

tzaeschke:
@ibauersachs The central part of our project is a new routing protocol (essentially replacing BGP), see https://www.scion.org/. I am working on a Java client, and there are clients in other languages (Golan, Rust, ...). In the end, many applications on a machine, including possibly the kernel/OS itself, should use the new protocol. We support using a separate hosts file in `/etc/scion/hosts`, but it would be good to keep using the `/etc/hosts`, otherwise it may become more or less useless for any addresses outside the local network (local autonomous system), and we end up with two separate configuration files to specify host addresses. 
To many, replacing BGP sounds like an impossible undertaking, but several ISPs (including British Telecom and all major Swiss ISPs) now support SCION. Also, the complete Swiss banking network has migrated to SCION. The reference implementations of all components are free open source owned by a non-profit organization.

PR #380, opened 2025-06-11
Add concurrency test for processReadyKeys() to detect potential thread-safety issues

bhaveshthakker:
This PR introduces a unit test that verifies the thread-safety of NioClient.processReadyKeys() under concurrent channel registration and readiness.

The test assumes correct behavior — i.e., no ConcurrentModificationException should be thrown when new channels become ready while processReadyKeys() is iterating over selector.selectedKeys(). If such an exception is thrown, the test fails, indicating a concurrency flaw in the current implementation.

Key points:

Uses a real selector loop via NioClient.selector(), with no changes to production logic.

Concurrently registers channels and makes them ready using Pipe.sink().write() to trigger select() updates.

Catches ConcurrentModificationException using a custom Thread.UncaughtExceptionHandler.

Includes processing delay to increase overlap and expose the race condition.

| Note: As this is a concurrency test, it may not fail consistently across runs or environments. Flakiness is expected until the underlying issue is fixed.

This test is intended as a reproducible regression check for an existing bug and will serve as a validation gate once a thread-safe fix is applied to processReadyKeys().

### DIFF
--- a/src/test/java/org/xbill/DNS/NioTcpClientTest.java
+++ b/src/test/java/org/xbill/DNS/NioTcpClientTest.java
@@ -17,6 +17,7 @@ import java.net.SocketTimeoutException;
 import java.nio.ByteBuffer;
 import java.time.Duration;
 import java.util.ArrayList;
+import java.util.ConcurrentModificationException;
 import java.util.List;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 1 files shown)


bhaveshthakker:
Yes, close() is the only place from issue seems to be reproducible!

Regarding the earlier fix (bhaveshthakker/dnsjava#1): that was created while I was experimenting with the failure, and I ended up pushing it prematurely. It wasn't intended as a finalized PR, so please feel free to disregard it.

Regarding unit test, yes, it was written just to reproduce the issue

Waiting for the solution to come in. LMK if I can be of any help

ibauersachs:
Does the linked PR #381 work for you?

PR #381, opened 2025-06-14
Run NIO close in the selector thread to prevent CME

### LINKED ISSUE #379: ConcurrentModificationException in NioClient.processReadyKeys() due to concurrent selector modifications
The NioClient.processReadyKeys() method is vulnerable to a java.util.ConcurrentModificationException. This occurs when another thread modifies the selector.selectedKeys() set while processReadyKeys() is iterating over it. This can happen when new channels are registered with the selector while processReadyKeys() is in progress.

**Steps to Reproduce:**

Create a scenario where new channels are registered with the selector while processReadyKeys() is running.
Ensure that the KeyProcessor takes some time to execute.
Trigger a high load of DNS requests.
A test case can be provided to reproduce the issue.

**Expected Behavior:**

The processReadyKeys() method should handle concurrent modifications to the selector.selectedKeys() set gracefully, without throwing a ConcurrentModificationException.

Exception in thread "dnsjava NIO selector" java.util.ConcurrentModificationException
at java.base/java.util.HashMap$HashIterator.nextNode(HashMap.java:1626)
at java.base/java.util.HashMap$KeyIterator.next(HashMap.java:1654)
at org.xbill.DNS.NioClient.processReadyKeys(NioClient.java:175)
at org.xbill.DNS.NioClient.runSelector(NioClient.java:130)
at java.base/java.lang.Thread.run(Thread.java:1520)

**Actual Behavior:**

The processReadyKeys() method throws a java.util.ConcurrentModificationException when another thread modifies the selector.selectedKeys() set during iteration.

**Proposed Solution:**

The recommended solution is to collect the keys to be removed in a separate collection (e.g., an ArrayList) and then remove them from the selector.selectedKeys() set after the iteration is complete.

**Additional Context:**

This issue can be difficult to reproduce because it depends on specific timing and concurrency conditions. However, it can cause instability and unexpected behavior in applications under load.

**Impact:**

This issue can cause application crashes and instability.

ibauersachs:
Closes #379
Closes #380

### DIFF
--- a/src/main/java/org/xbill/DNS/NioClient.java
+++ b/src/main/java/org/xbill/DNS/NioClient.java
@@ -10,6 +10,7 @@ import java.nio.channels.ClosedSelectorException;
 import java.nio.channels.SelectionKey;
 import java.nio.channels.Selector;
 import java.util.Iterator;
+import java.util.function.Consumer;
 import lombok.AccessLevel;
 import lombok.NoArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
@@ -24,9 +25,9 @@ import org.xbill.DNS.utils.hexdump;
  * <p>The following configuration parameter is available:
  *
  * <dl>
- *   <dt>dnsjava.nio.selector_timeout
+ *   <dt>{@value SELECTOR_TIMEOUT_PROPERTY}
  *   <dd>Set selector timeout in milliseconds. Default/Max 1000, Min 1.
- *   <dt>dnsjava.nio.register_shutdown_hook
+ *   <dt>{@value REGISTER_SHUTDOWN_HOOK_PROPERTY}
  *   <dd>Register Shutdown Hook termination of NIO. Default True.
  * </dl>
  *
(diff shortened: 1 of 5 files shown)


src/main/java/org/xbill/DNS/NioClient.java
```diff
@@ -81,6 +89,10 @@ public static void close() {
 
   private static void close(boolean fromHook) {
```
bhaveshthakker, line 90: Now, since the close() function is not actually closing the selector, can we rename this to  either of markForClose(), markForShutdown(), scheduleClose(), etc. Close() is misleading!

Also, now since the logic for close is not very sequentially straightforward, possibly we can add a comment over the function what this function does and how selector will read the flag later to actually close it

src/test/java/org/xbill/DNS/NioTcpClientTest.java
```diff
@@ -47,6 +56,136 @@ void testSelectorTimeoutLimits(int timeout) {
     }
   }
 
+  /**
```
bhaveshthakker, line 59: While we here, possibly we can add a similar test in NioUdpClientTest too for completeness

bhaveshthakker (commented):
Overall LGTM!

src/test/java/org/xbill/DNS/NioTcpClientTest.java
ibauersachs: Feel free to create a PR once this one is merged.

src/main/java/org/xbill/DNS/NioClient.java
bhaveshthakker: Ahh yes! closer() is public!! Well then adding a javadoc would suffice!

PR #399, opened 2026-02-20
Fix config option key names in README

lostnet:
The following references seem to have slightly different names
in the code than indicated:

dnsjava.dnssec.digest_enabled
dnsjava.dnssec.algorithm_enabled
org.jitsi.dnssec.nsec3.iterations

closes #398


ibauersachs:
Thanks!

PR #405, opened 2026-04-13
Add getXxxAsByteArray() getters and overload with escape boolean

### LINKED ISSUE #404: Access to "raw" strings in NAPTR record
Hello :)

I'm on a quest to parse NAPTR's `regexp` field as per [RFC 3402, section 3.2](https://datatracker.ietf.org/doc/html/rfc3402#section-3.2).

I'm partially stuck because the [`getRegexp`](https://github.com/dnsjava/dnsjava/blob/627813316446772d3736bf67a1493fb9e74bd26a/src/main/java/org/xbill/DNS/NAPTRRecord.java#L125) function converts the bytes to string via `byteArrayToString`: this adds back the escape character (`\`) where needed for string representation.

However, to correctly parse the `regexp` field, I would need the underlying unescaped value of this field.

For instance, if the regexp field was `!a\!b!c!i` I would have to parse:
- delim: `!`
- ere: `a!b`
- repl: `c`
- flags: `i`

However, in textual representation, `\` itself would have to be escaped, e.g.:
```kotlin
val naptr = Record.fromString(Name.fromConstantString("example.com"), Type.NAPTR, DClass.IN, """!a\\!b!c!i""", Name.root) as NAPTRRecord

naptr.getRegexp() // returns the string "!a\\!b!c!i"
```

There is currently no way to access the raw bytes of the `regexp` field. However, this is possible for TXT records using [`getStringsAsByteArrays`](https://github.com/dnsjava/dnsjava/blob/627813316446772d3736bf67a1493fb9e74bd26a/src/main/java/org/xbill/DNS/TXTBase.java#L110).

Would it be possible to introduce a `getRegexpAsByteArray` function that simply exposes the underlying `regexp` field? The same could be done for flags and service.

Would a PR that adds this be accepted?

MMauro94:
As discussed in the issue below, the library currently exposes character strings only as escaped strings for textual representation. This is a limitation as users are forced to deal with unnecessary escapes for application logic, and are also limited to interpret bytes as UTF-16 (as `byteArrayToString` casts bytes to `char`), whilst RFCs do not state which encoding for bytes should be used.

This commit adds:
- `getXxxAsByteArray()` getters that expose the "raw" byte array, allowing full control to users. This follows the already established pattern in `TXTBase` with the `getStringsAsByteArrays()` method
- an overload of the existing getters, which allows to pass an `escape` boolean: `true` (the default) is the current behavior, while `false` simply converts the bytes to a String using the UTF-8 encoding

Closes #404

### DIFF
--- a/src/main/java/org/xbill/DNS/CAARecord.java
+++ b/src/main/java/org/xbill/DNS/CAARecord.java
@@ -4,6 +4,7 @@
 package org.xbill.DNS;
 
 import java.io.IOException;
+import java.nio.charset.StandardCharsets;
 
 /**
  * Certification Authority Authorization
@@ -79,12 +80,32 @@ public class CAARecord extends Record {
 
   /** Returns the tag. */
   public String getTag() {
-    return byteArrayToString(tag, false);
+    return new String(tag, StandardCharsets.US_ASCII);
   }
 
-  /** Returns the value */
+  /**
+   * Returns the value as a string.
+   *
+   * @param escape if true, returns the RR textual representation of the underlying bytes. If false,
+   *     returns just the simple string using the UTF-8 charset with no additional escaping.
+   * @since 3.6.5
+   */
+  public String getValue(boolean escape) {
+    return escape ? byteArrayToString(value, false) : new String(value, StandardCharsets.UTF_8);
+  }
+
+  /** Returns the value as a string, escaped for RR textual representation */
   public String getValue() {
-    return byteArrayToString(value, false);
+    return getValue(true);
+  }
+
+  /**
+   * Returns the value as a raw byte-array
+   *
+   * @since 3.6.5
+   */
+  public byte[] getValueAsByteArray() {
+    return value;
   }
 
   @Override
--- a/src/main/java/org/xbill/DNS/NAPTRRecord.java
+++ b/src/main/java/org/xbill/DNS/NAPTRRecord.java
@@ -4,6 +4,7 @@
 package org.xbill.DNS;
 
 import java.io.IOException;
+import java.nio.charset.StandardCharsets;
 
 /**
  * Name Authority Pointer Record - specifies rewrite rule, that when applied to an existing string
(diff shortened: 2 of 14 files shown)


MMauro94:
Forgot to add: note that this has the unintended side-effect of exposing byte arrays outside of the record classes, which means that they could be changed outside of the class' control, hence making the class not fully immutable.

I assume that this is OK as AFAIK this library does not make any hard contracts about class mutability, especially as this pattern of exposing the byte array already exists in the aforementioned `getStringsAsByteArrays()`.

An alternative is to return a read-only [ByteBuffer](https://docs.oracle.com/javase/8/docs/api/java/nio/ByteBuffer.html) wrapping our original byte array (see [`wrap()`](https://docs.oracle.com/javase/8/docs/api/java/nio/ByteBuffer.html#wrap-byte:A-)). However, this would make it slightly less ergonomic as [there's no one-liner to convert it back to a byte array](https://stackoverflow.com/questions/28744096/convert-bytebuffer-to-byte-array-java) and also contains state about the position.

Another option is to copy the array each time: not great performance-wise, but data for a DNS record is quite limited anyway.

src/main/java/org/xbill/DNS/CAARecord.java
```diff
@@ -78,13 +79,29 @@ public int getFlags() {
   }
 
   /** Returns the tag. */
+  public String getTag(boolean escape) {
```
ibauersachs: As per https://datatracker.ietf.org/doc/html/rfc8659#section-4.1.1 the tag is "A non-zero-length sequence of ASCII letters and numbers in lowercase". So for this case, I don't think we should offer (non)-escaping.

src/main/java/org/xbill/DNS/NAPTRRecord.java
```diff
@@ -111,18 +112,42 @@ public int getPreference() {
   }
 
   /** Returns flags */
+  public String getFlags(boolean escape) {
```
ibauersachs: https://datatracker.ietf.org/doc/html/rfc3403#section-4
```
FLAGS
A <character-string> containing flags to control aspects of the
rewriting and interpretation of the fields in the record.  Flags
are single characters from the set A-Z and 0-9.  The case of the
alphabetic characters is not significant.
```

I don't think the escape option (and corresponding) getFlagsAsByteArray makes sense here

src/main/java/org/xbill/DNS/NSAPRecord.java
```diff
@@ -79,8 +80,16 @@ protected void rdataFromString(Tokenizer st, Name origin) throws IOException {
   }
 
   /** Returns the NSAP address. */
+  public String getAddress(boolean escape) {
```
ibauersachs: https://datatracker.ietf.org/doc/html/rfc1706 specifies the address as RDATA. Presenting this as a string seems entirely wrong, thus:
- no new getAddress(bool)
- mark getAddress() as obsolete, pointing to getAddressAsByteArray as replacement

src/main/java/org/xbill/DNS/URIRecord.java
```diff
@@ -79,8 +80,16 @@ public int getWeight() {
   }
 
   /** Returns the target URI */
+  public String getTarget(boolean escape) {
```
ibauersachs: https://datatracker.ietf.org/doc/html/rfc7553#section-4.4

Target is not a `<character string>`. How about instead of getTargetAsByteArray add a `java.net.URI getTargetAsUri()` and mark getTarget obsolete?

src/main/java/org/xbill/DNS/X25Record.java
```diff
@@ -59,8 +60,16 @@ protected void rdataFromString(Tokenizer st, Name origin) throws IOException {
   }
 
   /** Returns the X.25 PSDN address. */
+  public String getAddress(boolean escape) {
```
ibauersachs: https://datatracker.ietf.org/doc/html/rfc1183#section-3.1

An X25 address consists only of digits (and is validated in this class). An escape option doesn't make sense here.

ibauersachs (changes requested):
Thanks!

Besides the inline comments, please:
- Add a proper Javadoc to all new or changed methods, documenting the parameter 
- Mark all new methods with `@since 3.6.5`

src/main/java/org/xbill/DNS/CAARecord.java
MMauro94: Good point. Given the RFC specifies ASCII, we can just keep the string return. Given the set of allowed characters `[0-9a-zA-Z]` doesn't intersect with any character that would be escaped by `byteArrayToString`, I've replaced the return with `new String(value, StandardCharsets.US_ASCII)`.

MMauro94:
Sorry it took a while, I've pushed the changes requested in the comments.

ibauersachs (approved):

ibauersachs:
> Sorry it took a while, I've pushed the changes requested in the comments.

No worries. Thanks a lot!

Issue #129, opened 2020-09-14
java.nio.channels.CancelledKeyException


RoyJing:
After our app was released, we had a lot of crashes like this. The SDK version is 3.0.2, this detailed stack information

Fatal Exception: java.nio.channels.CancelledKeyException
       at sun.nio.ch.SelectionKeyImpl.ensureValid(SelectionKeyImpl.java:73)
       at sun.nio.ch.SelectionKeyImpl.readyOps(SelectionKeyImpl.java:87)
       at java.nio.channels.SelectionKey.isReadable(SelectionKey.java:289)
       at org.xbill.DNS.NioUdpClient$Transaction.processReadyKey(NioUdpClient.java:119)
       at org.xbill.DNS.Client.processReadyKeys(Client.java:79)
       at org.xbill.DNS.Client.runSelector(Client.java:62)
       at org.xbill.DNS.Client.lambda$jAWiqsJXxSKrE-3ZAhrF3bfbwqQ(Client.java)
       at o.-$$Lambda$fnu$jAWiqsJXxSKrE-3ZAhrF3bfbwqQ.run(-.java)
       at java.lang.Thread.run(Thread.java:764)

ibauersachs:
The current version is 3.2.2. I cannot support older versions and the particular issue was AFAIK fixed in 3.1.0.

RoyJing:
> The current version is 3.2.2. I cannot support older versions and the particular issue was AFAIK fixed in 3.1.0.

Great, I'll upgrade to the latest version

ibauersachs:
@AlexanderGH Do you have more information when this happens? I found an old JDK bug (https://bugs.openjdk.org/browse/JDK-4729342) which says that "old" selection keys should be cancelled. We don't do this from what I can tell at a brief glance.
It would help tremendously if you could come up with a reproducer, even if it only happens in 1 out of 10 runs.

AlexanderGH:
So sadly this only happens in an Android app in prod. As best we can tell, this correlates to when we System.exit(0) or possibly restart the app in the background after System.exit(0)ing.

Stats:
- It happens on Android versions all the way up to 16, so if it is a bug in ART, it hasn't been fixed.
- Doesn't look like it's caused by an OEM modification since it happens on many OEMs

ibauersachs:
The exit part at least correlates with what I'd assume could happen as a race condition. Initially I thought this was during normal lookups while the application was running.

Can you test if the following patch helps?

```patch
[...]

PR #411, opened 2026-07-10
Handle canceled keys in TCP client

techee:
**Disclaimer**: the whole patch, including the description below, was generated using Claude Fable using the xhigh effort - I only experienced the crash on Android and asked it to fix the problem, but I didn't dig into the problem myself. Please let me know if the patch needs any adjustments.

---

## Problem

A failed write in `NioTcpClient.ChannelState#processWrite` (e.g. the server resetting the
connection while multiplexed queries are still being written) kills the shared selector
thread with an unhandled `CancelledKeyException`:

```
java.nio.channels.CancelledKeyException
    at sun.nio.ch.SelectionKeyImpl.ensureValid(SelectionKeyImpl.java:73)
    at sun.nio.ch.SelectionKeyImpl.interestOps(SelectionKeyImpl.java:82)
    at org.xbill.DNS.NioTcpClient$ChannelState.processWrite(NioTcpClient.java:282)
    at org.xbill.DNS.NioTcpClient$ChannelState.processReadyKey(NioTcpClient.java:148)
    at org.xbill.DNS.NioClient.processReadyKeys(NioClient.java:255)
    at org.xbill.DNS.NioClient.runSelector(NioClient.java:156)
```

There are two escape paths:

1. `processWrite` catches the `IOException` from `Transaction#send`, calls `key.cancel()`,
   and then falls through to `key.interestOps(SelectionKey.OP_READ)` on the canceled key.
2. Even with 1 fixed, `processReadyKey` calls `key.isReadable()` right after `processWrite`
   returns; on the now-canceled key this also throws (`isReadable` → `readyOps` →
   `ensureValid`).

Since `runSelector` only handles `IOException`/`ClosedSelectorException`, the exception
kills the selector thread. All pending and future queries then never complete — query
timeouts run on that same thread — and on Android, where an uncaught exception on any
thread is fatal, the whole application crashes. We see this regularly in the field from an
app whose DNS tool issues parallel queries with DNSSEC (truncation → TCP fallback,
multiplexed on one channel) against home routers that reset the TCP connection mid-write.

This is the TCP counterpart of #129 ("Handle canceled keys in UDP client", fixed in 3.6.5).

## Fix

- `processWrite`: treat a write failure like the other error paths in this class
  (`processConnect`, `processRead`) — a failed write leaves the TCP stream in an undefined
  state for every transaction multiplexed on the channel, so fail them all via
  `handleChannelException`, cancel the key, and return without touching `interestOps`.
- `processReadyKey`: re-check `key.isValid()` before the `isReadable()` branch, since
  `processWrite` may have canceled the key.

## Test

`testWriteFailureDoesNotKillSelectorThread` reproduces the crash deterministically: large
queries against a server with a 16-byte receive buffer (the `testResponseStream` trick)
leave transactions queued mid-write with `OP_WRITE` armed; the server then resets the
connection, so the next write fails inside `processWrite`. The test asserts that all
transactions complete exceptionally and that the selector thread survives (watched via an
uncaught-exception handler, the same pattern as
`testProcessReadyKeysShouldNotThrowConcurrentModificationException`).

Without the fix the test fails deterministically (selector thread dies, transactions never
complete); with the fix, the whole `NioTcpClientTest`/`NioUdpClientTest` suites pass.


techee:
This is the text from claude code for more context:  

What the PR contains (commit 0ce1ec8, "Handle canceled keys in TCP client")

  The fix turned out to need two changes, not one — and the test proved it:

  1. processWrite: a failed write now routes through handleChannelException —
  failing every transaction multiplexed on the channel and closing it, matching
  how processConnect/processRead already handle errors — then cancels the key
  and returns, instead of falling through to interestOps() on the cancelled key.
  2. processReadyKey: a validity re-check before the isReadable() branch. This
  one I only discovered because my first test run failed with fix 1 applied:
  isReadable() internally calls ensureValid(), so it throws the very same
  CancelledKeyException one line after processWrite returns. A one-line fix
  would have shipped the same crash through a second door.

  Plus a regression test, testWriteFailureDoesNotKillSelectorThread, built from
  the repo's own idioms: it fills the client's send buffer against a server with
  a 16-byte receive buffer (their testResponseStream trick) so transactions are
  stuck mid-write, then RSTs the connection — hitting the write-failure path
  deterministically. It asserts the transactions fail cleanly and that the
  selector thread survives a grace period, because I caught my first version
  false-passing when its assertions raced the thread's death.

  Verification, both directions: without the fix the test fails
  deterministically (selector thread dies, transactions hang — your production
  symptom exactly); with it, the full NioTcpClientTest + NioUdpClientTest suites
  pass (12/12), spotless-formatted. I used a self-contained Maven in
  ~/Downloads/apache-maven-3.9.16 with the Android Studio JDK — no system
  changes.

  ★ Insight ─────────────────────────────────────
  - The repo's history was the design guide: the maintainer fixed the identical
  bug in the UDP client four months ago (#129, released in 3.6.5) with exactly
  this shape — channel-level failure handling, key.cancel() + return, and a
  selector-thread-watching test. The TCP side was simply left behind, which
  makes this PR an easy "complete the symmetry" review.
  - The red-test-first discipline paid for itself twice here: it exposed the
  second escape path (isReadable → ensureValid) that code inspection missed, and
  it exposed the test's own race. A patch verified only by reading would have
  been wrong in both ways.
  ─────────────────────────────────────────────────

techee:
Forgot to mention - feel free to take over the patch and commit it yourself. I'm not its author anyway so no credits needed.

ibauersachs:
Thanks!

PR #413, opened 2026-08-26
hosts: resolve PTR queries from the local hosts database

### LINKED ISSUE #350: Support parsing PTR from /etc/hosts
It doesn't seem to support it now, do we need support?

ChrisJr404:
Closes #350. HostsFileParser can now resolve a reverse (PTR) lookup back to the first host name for an address, and both Lookup and LookupSession consult it for PTR queries the same way they already do for A and AAAA. Added unit tests for the parser and integration tests for both lookup paths.


ibauersachs:
Thanks!

